PR #3, opened 2020-10-28
Add more CRUD methods to the iOS implementation

migbot:
#### Summary
* Adds put, patch, and delete
* Adds SwiftyJSON for easier parsing of body to be used as AF parameters
* Includes a few example app fixes


shazm (approved):
I can't comment too much on the iOS code but it builds and runs fine for me :)

example/src/screens/NetworkClientScreen.tsx
```diff
@@ -161,21 +161,25 @@ export default function NetworkClientScreen({navigation, route}) {
...
-        switch (method.toLowerCase().trim()) {
-            case 'get':
-                requestMethod = client.get;
-                break;
-            case 'post':
-                requestMethod = client.post;
-                break;
-            default:
-                break;
+        const methodName = method.toLowerCase().trim();
```
enahum, line 165: non blocking, but here you are trimming the method name and applying a lowercase to compare, then on  line 374 is totally different.

enahum (approved):
For a different PR please consider adding a method to remove/delete a client (think of when a server is removed / logged out)

PR #4, opened 2020-10-29
Updates project to use strict type definitions

shazm:
Includes various changes:
- `types` directory holds type definitions
- Extends React Native Module types for `NetworkClient`
- Removes eslint rule overrides
- Adds example and lib directory to eslint and typescript ignore (amongst others)

This PR should also allow all the scripts to run as well (`npm run prepare` etc), with ESLint / Prettier kicking in for auto-formatting and so forth as well.


enahum:
before looking at the PR,  can you replace the `@types` directory for `types` I worry that having the folder starting with `@` it can conflict if we ever need to add a type definition dependency

enahum (changes requested):
see https://github.com/mattermost/react-native-network-client/pull/4#issuecomment-718637733

.eslintrc.json
```diff
@@ -22,30 +22,30 @@
...
-          "vars": "all",
-          "args": "after-used"
-        }
-      ],
-      "@typescript-eslint/no-explicit-any": "warn",
-      "@typescript-eslint/no-use-before-define": 0,
-      "@typescript-eslint/no-var-requires": 0,
-      "@typescript-eslint/explicit-function-return-type": 0,
-      "@typescript-eslint/explicit-module-boundary-types": "off"
+      // "global-require": 0,
```
enahum, line 25: why disable the rules?

@types/ApiClient.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    headers?: Record<string, string>;
+    body?: Record<string, string>;
+}
+
+type Headers = Record<string, string>;
+
+type Response = {
+    headers?: Headers;
+    data?: Record<string, string>;
+}
```
enahum, line 14: headers: ok -> not optional
data: ok? -> this could also be an array, plain text or a blob -> maybe consider adding some functions like base64(), json(), text() to conform with the fetch API
status: missing
redirects: missing -> this should be an array with all url's

@types/ApiClient.d.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface GenericClientInterface {
+    get(endpoint: string, options?: RequestOptions): Promise<Response>;
```
enahum, line 17: wondering why the `GenericClientInterface` can only use the `GET` method

@types/ApiClient.d.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type RequestOptions = {
+    headers?: Record<string, string>;
+    body?: Record<string, string>;
```
enahum, line 6:
```suggestion
    body?: Record<string, string> | string | Array<unknown>;
```

other options:
* backgroundTask
* timeout
* followRedirects (default to true)
*  maxRetries
* backoffType
* backoffInterval

@types/NativeNetworkClient.d.ts
```diff
@@ -0,0 +1,16 @@
...
+interface NativeNetworkClient {
+    get(baseUrl: string, endpoint: string | null, options?: RequestOptions): Promise<Response>;
+    put(baseUrl: string, endpoint: string | null, options?: RequestOptions): Promise<Response>;
+    post(baseUrl: string, endpoint: string | null, options?: RequestOptions): Promise<Response>;
+    patch(baseUrl: string, endpoint: string | null, options?: RequestOptions): Promise<Response>;
+    delete(baseUrl: string, endpoint: string | null, options?: RequestOptions): Promise<Response>;
+
+    getApiClientHeadersFor(baseUrl: string): Promise<Headers>;
+    addApiClientHeadersFor(baseUrl: string, headers: Headers): Promise<void>;
+    createApiClientFor(baseUrl: string, config?: ApiClientConfiguration): Promise<void>;
```
enahum, line 12: we should add `removeApiCientFor` and `removeWebSocketFor`

enahum (changes requested):

@types/ApiClient.d.ts
migbot: No reason. We can add all methods there in some future PR.

migbot: Fixed: https://github.com/mattermost/react-native-network-client/pull/5

@types/ApiClient.d.ts
migbot: https://mattermost.atlassian.net/browse/MM-30108

@types/ApiClient.d.ts
migbot: Background tasks: https://mattermost.atlassian.net/browse/MM-30109
Timeout: https://mattermost.atlassian.net/browse/MM-30114
Follow Redirects: https://mattermost.atlassian.net/browse/MM-30118
Retries: https://mattermost.atlassian.net/browse/MM-22753

@types/NativeNetworkClient.d.ts
migbot: https://github.com/mattermost/react-native-network-client/pull/6

.eslintrc.json
shazm: Most of these rules were actually disabling a lot of the recommend checks from the esling/typescript/mattermost-react rules, so throwing lots of warnings and impeding development. The PR sets up the types so we can adhere to all the recommended standards and not need to have rules disabling various checks / warnings.

migbot (approved):

CONTRIBUTING.md
```diff
@@ -3,3 +3,87 @@
 Thank you for your interest in contributing! Please see the [Mattermost Contribution Guide](https://developers.mattermost.com/contribute/getting-started/) which describes the process for making code contributions across Mattermost projects and [join our "Native Mobile Apps" community channel](https://pre-release.mattermost.com/core/channels/native-mobile-apps) to ask questions from community members and the Mattermost core team.
 
 When you submit a pull request, it goes through a [code review process outlined here](https://developers.mattermost.com/contribute/getting-started/code-review/).
+
+## Development workflow
+
+To get started with the project, run `yarn bootstrap` in the root directory to install the required dependencies for each package:
+
+```sh
+yarn bootstrap
```
enahum, line 12: not blocking, but lets change all yarn references to use npm

src/ApiClient/index.tsx
```diff
@@ -4,48 +4,40 @@
...
-      this.baseUrl = baseUrl;
-  }
+    constructor(baseUrl: string) {
+        this.baseUrl = baseUrl;
+    }
 
-  getHeaders = (): Promise<object> => NetworkClient.getApiClientHeadersFor(this.baseUrl);
-  addHeaders = (headers: object): void => NetworkClient.addApiClientHeadersFor(this.baseUrl, headers);
+    getHeaders = (): Promise<Headers> => NetworkClient.getApiClientHeadersFor(this.baseUrl);
+    addHeaders = (headers: Headers): Promise<void> => NetworkClient.addApiClientHeadersFor(this.baseUrl, headers);
```
enahum, line 27: not blocking, should we consider returning something here? could it be that we are trying to add a header and that could fail? if the answer is no, then nvm

types/ApiClient.d.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type RequestOptions = {
+    headers?: Record<string, string>;
+    body?: Record<string, string>;
```
enahum, line 6: this could be a string or and array too, maybe a separate ticket?. One good exercise would be to check all request options that we have at https://api.mattermost.com

enahum (approved):
Please add tickets or address my comments before merging.

src/ApiClient/index.tsx
shazm: I think eventually yes; it'd be good to return the header objects to prevent subsequent calls - but for now the promise either resolves or rejects (throws errors) so we do get confirmation and that should be okay.

types/ApiClient.d.ts
shazm: Good point! I've created https://mattermost.atlassian.net/browse/MM-30180

PR #6, opened 2020-10-29
[MM-30125] JS API: Add functions to remove clients

migbot:
#### Summary
* Adds `invalidate` client methods
* Exports two separate native modules, `GenericClient` and `APIClient`
* Rejects on native side if client has been invalidated
* `getOrCreate*` methods return `created` boolean

Note: I'll wait for the type changes to be merged to master and fix the conflicts in this PR prior to merging.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30125


shazm (approved):
Is the implementation for them coming separately? I think the iOS side is missing the `RCT_EXTERN_METHOD`, and it'd be good to delete sessions off the list in the example app.

enahum (commented):
LGTM, 100% agree with @shazm

src/ApiClient/index.tsx
```diff
@@ -34,15 +44,20 @@ class ApiClient implements ApiClientInterface {
   getHeaders = (): Promise<object> => NetworkClient.getApiClientHeadersFor(this.baseUrl);
   addHeaders = (headers: object): void => NetworkClient.addApiClientHeadersFor(this.baseUrl, headers);
 
-  get = (endpoint: string, options?: RequestOptions): Promise<Response> => NetworkClient.get(this.baseUrl, endpoint, options);
+  get = (endpoint: string, options?: RequestOptions): Promise<Response> => {
+    if (!CLIENTS[this.baseUrl]) {
```
migbot, line 48: @shazm @enahum What do you think about a this? I don't really like having to check `CLIENTS` in each of the ApiClient class methods...

An alternative is to let the call go through and have the native side reject if the session does not exist, however, we'll need separate native methods for the GenericClient to call since I'm currently defaulting to using the [default session when a session is not found](https://github.com/mattermost/react-native-network-client/blob/703b5abafeae7e1fd32cc3a024258afe98907c47/ios/NetworkClient.swift#L52).

enahum (approved):
LGTM

PR #8, opened 2020-10-31
chore: various project fixes and dep updates

enahum:
#### Summary
Multiple project fixes:
* Android & node package name
*  Upgrade to RN 63.3 including LaunchScreen for example app that follows OS Appearance
* CONTRIBUTING commands
* Enabled Hermes for example app to help build the lib hermes compatible


shazm (approved):
All looks good to me, and thanks for fixing my copy/paste fail in the contributing doc!

Quick question - is there a reason we peg version numbers and not allow the minor / dot version increments? Just from the changes in package.json; removal of the preceding `^` in version numbers.

enahum:
We definitely prefer fixed versions, that way we prevent surprises

migbot (approved):

PR #9, opened 2020-11-11
[MM-30119] [MM-30120] Support follow redirects for iOS

migbot:
#### Summary
* Adds configuration for follow redirects
* Uses Alamofire's `Redirector` for redirect handling, but only supports `.follow` and `.doNotFollow` because have no use for `.modify` at Mattermost. Also, it's not clear how we can drive the modify closure from the JS side.
* Adds boolean property `followRedirects` to the API client class
* Returns `lastRequestedUrl` as part of the response. RNFetchBlob includes an array of redirect URLs, however, this was difficult to implement on the iOS side. I tried using stored properties in extensions of `URLSessionTask` and `URLRequest` to keep track of each URL from the redirector but wasn't successful. In the end, we only ever use the last requested URL from the RNFetchBlob array and that is accessible in the Alamofire response:`json.response.url`. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30119
https://mattermost.atlassian.net/browse/MM-30120


example/package.json
```diff
@@ -9,6 +9,7 @@
     "start": "react-native start"
   },
   "dependencies": {
+    "@react-native-community/checkbox": "^0.5.5",
```
enahum, line 12: use exact

ios/NetworkClient.swift
```diff
@@ -109,21 +112,29 @@ class NetworkClient: NSObject {
...
+        if let headers = RCTConvert.nsDictionary(options["headers"]) {
+            config.httpAdditionalHeaders = headers
         }
         
         return config
     }
+
+    func getRedirectHandlerFrom(options: Dictionary<String, Any>) -> RedirectHandler? {
+        if let redirectHandlerConfig = RCTConvert.nsDictionary(options["redirectHandlerConfig"])  {
+            return Redirector(behavior: redirectHandlerConfig["follow"] as! Bool ? .follow : .doNotFollow)
```
enahum, line 133: 💯

enahum (approved):

enahum (approved):
lgtm

enahum:
@shazm please review

PR #10, opened 2020-11-12
[MM-30115] [MM-30116] Add URLSessionConfiguration and URLRequest config options

migbot:
#### Summary
Note: Branched out from https://github.com/mattermost/react-native-network-client/pull/9; will merge after that PR is merged. Changes for tickets below are in commits https://github.com/mattermost/react-native-network-client/pull/10/commits/6501d21d7d623ca981b73099e3ce898a3702e3d4 and https://github.com/mattermost/react-native-network-client/pull/10/commits/0f2253cfc0d5dc87923b5816bf528cc9a8d7b61d

* Adds support for setting `URLSessionConfiguration` options `allowsCellularAccess`, `waitsForConnectivity`, `timeoutIntervalForRequest`, `timeoutIntervalForResource`, and `httpMaximumConnectionsPerHost`
* Adds support for setting `URLRequest` option `timeoutInterval`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30115
https://mattermost.atlassian.net/browse/MM-30116

PR #12, opened 2020-11-17
[MM-30122][MM-30123] iOS support for retries

migbot:
#### Summary
* Adds iOS support for retries with exponential backoff for the URLSessionConfiguration and per request
* Allows cancellation of all requests on 401
* Updated GenericClient to handle timeout interval for requests as well (this was missed in https://github.com/mattermost/react-native-network-client/pull/10)

Note: There's quite a bit of code duplication, both in the iOS code as well as the example app. This will be addressed in a separate PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30122
https://mattermost.atlassian.net/browse/MM-30123


enahum:
@migbot i'm ok with that, but want to be consistent, do we'll need the same values on Android

ios/APIClient.swift
```diff
@@ -10,16 +10,21 @@
 import Alamofire
 import SwiftyJSON
 
+let EXPONENTIAL_RETRY = "exponential"
```
enahum, line 13: what's the use of this constant? I saw it in the example file, can this be constant be exposed by the module ?

ios/APIClient.swift
```diff
@@ -10,16 +10,21 @@
...
     
     @objc(createClientFor:withOptions:withResolver:withRejecter:)
     func createClientFor(baseUrl: String, options: Dictionary<String, Any>?, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) -> Void {
-        if let options = options {
+        let options = JSON(options)
+        if options != JSON.null {
             let configuration = getURLSessionConfigurationFrom(options: options)
             let redirectHandler = getRedirectHandlerFrom(options: options)
+            let interceptor = getInterceptorFrom(options: options)
+            let cancelRequestsOnUnauthorized = options["sessionConfiguration"]["cancelRequestsOnUnauthorized"].boolValue
```
enahum, line 25: can `options["sessionConfiguration"]` be nil?

ios/APIClient.swift
```diff
@@ -99,45 +110,86 @@ class APIClient: NSObject {
...
+        let adapters = [RequestAdapter]()
+        
+        return adapters
+    }
+
+    func getRequestRetriersFrom(options: JSON) -> [RequestRetrier] {
+        var retriers = [RequestRetrier]()
+
+        let configuration = options["retryPolicyConfiguration"]
+        if configuration["type"].string == EXPONENTIAL_RETRY {
```
enahum, line 182: what is the other type of retries if no exponential ?

ios/GenericClient.swift
```diff
@@ -15,67 +15,101 @@ class GenericClient: NSObject {
...
+            let retryLimit = configuration["retryLimit"].uInt ?? RetryPolicy.defaultRetryLimit
+            let exponentialBackoffBase = configuration["exponentialBackoffBase"].uInt ?? RetryPolicy.defaultExponentialBackoffBase
+            let exponentialBackoffScale = configuration["exponentialBackoffScale"].double ?? RetryPolicy.defaultExponentialBackoffScale
+
+            let retryPolicy = RetryPolicy(retryLimit: retryLimit, exponentialBackoffBase: exponentialBackoffBase, exponentialBackoffScale: exponentialBackoffScale)
+            retriers.append(retryPolicy)
+        }
+
+        return retriers
+    }
```
enahum, line 95: this code is being duplicated, can you extract it to a different file or something ?

migbot: Donzo.

ios/APIClient.swift
migbot: Maybe at some point we (or someone for the community) wants to add some other retry algorigthm.

ios/APIClient.swift
migbot: Yes, but SwiftyJSON doesn't complain and will return `nil` anyways.

ios/APIClient.swift
migbot: Donzo.

shazm:
From the comments on the [Auth Token / Session PR](https://github.com/mattermost/react-native-network-client/pull/15); do we want to remove the RNNC prefix from our Native modules?

@types/ReactNative.ts
```diff
@@ -5,5 +5,6 @@ declare module "react-native" {
     export interface NativeModulesStatic {
         APIClient: NativeAPIClient;
         GenericClient: NativeGenericClient;
+        RNNCConstants: NativeConstants;
```
enahum, line 8: should we rename this to `NetworkConstants` ? 1/5

example/ios/Podfile
```diff
@@ -18,4 +18,5 @@ target 'NetworkClientExample' do
 
   # Pods for NetworkClientExample
   pod 'react-native-network-client', :path => '../..'
+  pod 'RNVectorIcons', :path => '../node_modules/react-native-vector-icons'
```
enahum, line 21: is this needed?

example/package.json
```diff
@@ -19,6 +19,7 @@
     "react-native-dropdown-picker": "3.7.1",
     "react-native-gesture-handler": "1.8.0",
     "react-native-modal": "11.5.6",
+    "react-native-numeric-input": "^1.8.3",
```
enahum, line 22:
```suggestion
    "react-native-numeric-input": "1.8.3",
```

ios/NetworkClient.swift
```diff
@@ -0,0 +1,130 @@
...
+            return Interceptor(adapters: adapters)
+        }
+        if (!retriers.isEmpty) {
+            return Interceptor(retriers: retriers)
+        }
+        
+        return nil
+    }
+
+    func getRequestAdapters(from options: JSON) -> [RequestAdapter] {
```
enahum, line 84: is this function complete?

enahum (approved):

enahum:
non-blocking comments

ios/NetworkClient.swift
migbot: Not yet, but needed to add it here because there's no init func for Interceptor where you just pass retriers.

example/ios/Podfile
migbot: Yes for the numeric input library (which I may remove later).

@types/ReactNative.ts
migbot: In the next PR.

PR #13, opened 2020-11-17
Type safety

shazm:
All Typescript and ESLint errors / warnings have been cleared, and type definitions live in a `.ts` file.

@enahum - I've moved the ts files into a `@types` folder, which seems to work fine and doesn't cause issues with `bob build`.


example/ios/Podfile.lock
```diff
@@ -381,20 +381,19 @@ DEPENDENCIES:
...
+    - Flipper
     - Flipper-DoubleConversion
     - Flipper-Folly
     - Flipper-Glog
     - Flipper-PeerTalk
     - Flipper-RSocket
-    - OpenSSL-Universal
-  trunk:
-    - CocoaAsyncSocket
-    - Flipper
```
shazm, line 396: I'm not sure about these - I had to run `pod install` to get the app up in iOS (@migbot

@types/APIClient.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+type ClientResponse = {
+    headers?: ClientHeaders;
+    data?: Record<string, unknown>;
+    code: number;
+    lastRequestedUrl: string;
+};
+
+interface GenericClientInterface {
```
migbot, line 19: We won't be configuring the generic client so no need for anything other than the CRUD methods. Headers can be added on a per request basis.

@types/NativeNetworkClient.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+interface NativeNetworkClient {
```
migbot, line 3: Where is this type used?

migbot (commented):
Sweet, no more `--no-verify` commits :)
Just a couple of comments below.

example/ios/Podfile.lock
migbot: Hmm, odd. Can you delete `Podfile.lock` and re-run `pod install`, then commit the new lock file?

@types/APIClient.ts
shazm: I had to add them in optionally as there was some code in the GenericClientScreen looking for those params.

@types/NativeNetworkClient.ts
shazm: It provides type safety when we do 

`const {NetworkClient} = NativeModules;`

via [this file](https://github.com/mattermost/react-native-network-client/blob/type-fixes/%40types/ReactNative.ts).

@types/APIClient.ts
migbot: Hmm, looking at [GenericClientScreen](https://github.com/mattermost/react-native-network-client/blob/289624c6ced9c1ca597e575d161aec9c03cc9793/example/src/screens/GenericClientScreen.tsx) and I don't see where it would need the extra methods. Is it maybe because of [options being of type `APIClientConfiguration`](https://github.com/mattermost/react-native-network-client/blob/289624c6ced9c1ca597e575d161aec9c03cc9793/example/src/screens/GenericClientScreen.tsx#L152)? Either way, those options should be of type `RequestOptions` no?

@types/NativeNetworkClient.ts
migbot: OK, I'm still unclear as to why the CRUD methods are in this interface. Shouldn't this interface only include the exported function type defs?

migbot: This makes total sense to me now 🤦, but since we're now exporting two modules do we need two of these?

@types/NativeGenericClient.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+interface NativeGenericClient {
+    get(baseUrl: string, options?: RequestOptions): Promise<ClientResponse>;
```
migbot, line 4:
```suggestion
    get(url: string, options?: RequestOptions): Promise<ClientResponse>;
```

example/src/screens/GenericClientScreen.tsx
```diff
@@ -129,29 +131,25 @@ const RequestHeader = ({index, header, updateHeader}) => {
...
+    const sanitizeHeaders = (headersArray: {key: string, value: string}[]) => {
+        const headers = headersArray
+            .filter( (k,v) => k && v)
+            .reduce( (prev, cur) => prev[cur.key] = cur.value, {} as any)
         return headers;
     }
 
     const makeRequest = async () => {
-        const options = {
+        const options: APIClientConfiguration = {
```
migbot, line 152:
```suggestion
        const options: RequestOptions = {
```

@types/APIClient.ts
migbot: @shazm No need for `baseUrl`, `config`, `getHeaders`, `addHeaders`, and `invalidate` for the GenericClient, right?

migbot (changes requested):

@types/APIClient.ts
shazm: Yup! Fixed :)

shazm:
Committed the suggestions and resolved feedback 👍

migbot (approved):

@types/APIClients.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type ClientHeaders = Record<string, string>;
+
+type RequestOptions = {
+    headers?: ClientHeaders;
+    body?: Record<string, unknown> | string;
```
enahum, line 8: This can be an array of something, we can add it later

@types/APIClients.ts
```diff
@@ -0,0 +1,65 @@
...
+    timeoutInterval?: number;
+};
+
+type ClientResponse = {
+    headers?: ClientHeaders;
+    data?: Record<string, unknown>;
```
enahum, line 14: Same here, not necessarily an object

@types/ReactNative.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+export * from "react-native";
+declare module "react-native" {
```
enahum, line 4: Why react-native? Shouldn't this be our own module?

enahum (approved):
I won't block it but consider making the changes

@types/ReactNative.ts
shazm: This essentially extends react-native's NativeModule; so we can import our APIClient and GenericClient with types / autocomplete / IDE goodness.

PR #15, opened 2020-12-02
[MM-30954] [MM-30955] iOS support for native level bearer token authentication for session requests 

migbot:
#### Summary
This is based on https://github.com/mattermost/react-native-network-client/pull/12 so the relevant changes for this PR are in commit https://github.com/mattermost/react-native-network-client/commit/7245d285485c1cfc6c47dfb43c7ca9777536701f

* Did not add native level basic authentication support as it's not used at Mattermost
* Did not need to do anything to support cookie based authentication. Cookies received from the server are automatically set in `session.configuration.httpCookieStorage` and sent with all session requests.
* For bearer authentication, you set the response header from which to extract the token in `requestAdapterConfiguration.bearerAuthTokenResponseHeader`. If this is set, a request adapter is configured to read the token from the Keychain, if it exists. The token is set in the Keychain when handling each response by extracting the token from a matching `bearerAuthTokenResponseHeader` of the response headers. The token is removed from the Keychain when the session is invalidated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30954
https://mattermost.atlassian.net/browse/MM-30955


@types/ReactNative.ts
```diff
@@ -5,5 +5,6 @@ declare module "react-native" {
     export interface NativeModulesStatic {
         APIClient: NativeAPIClient;
         GenericClient: NativeGenericClient;
+        RNNCConstants: NativeConstants;
```
shazm, line 8: I was thinking about this the other day - it's probably worth just calling this `Constants` to keep it inline with the naming above it no? `APIClient >> NativeAPIClient`, `Constants >> NativeConstants`

example/src/screens/APIClientScreen.tsx
```diff
@@ -147,17 +148,47 @@ const RequestHeader = ({ index, header, updateHeader }: RequestHeaderProps) => {
...
     const { client } = route.params;
 
-    const [method, setMethod] = useState("GET");
-    const [endpoint, setEndpoint] = useState("/");
-    const [timeoutInterval, setTimeoutInterval] = useState("");
-    const [body, setBody] = useState("");
+    const [method, setMethod] = useState(METHOD.POST);
+    const [endpoint, setEndpoint] = useState("/api/v4/users/login");
+    const [timeoutInterval, setTimeoutInterval] = useState(30);
+    const [body, setBody] = useState('{"login_id":"","password":""}');
```
shazm, line 154: I'm wondering if it's worth also setting a "default" `APIClient` that has MatterMost API details prefilled, vs on the creator here? So when you open the example app - you see a generic client, a mattermost api client and can add a client with blanks.

example/src/screens/APIClientScreen.tsx
```diff
@@ -355,14 +384,87 @@ export default function APIClientScreen({ route }: APIClientScreenProps) {
...
+                            />
+                        </View>
+                    </View>
+                </>
+            );
+        }
+
+        return (
+            <>
+                {checkbox}
```
shazm, line 454: I think you can also do `<Checkbox />` here as well which might be more readable; implying that it is a component that will be rendered

example/src/screens/GenericClientScreen.tsx
```diff
@@ -142,15 +145,46 @@ export default function GenericClientScreen({
...
+    const [retryPolicyConfiguration, setRetryPolicyConfiguration] = useState<
+        RetryPolicyConfiguration
+    >({
+        type: undefined,
+        retryLimit: 2,
+        exponentialBackoffBase: 2,
+        exponentialBackoffScale: 0.5,
+    });
+
     const scrollView = useRef<ScrollView>(null);
```
shazm, line 165: Just based on the last comment, it might be worth being explicit here calling it a `scrollViewRef`.

example/src/screens/GenericClientScreen.tsx
```diff
@@ -288,6 +324,79 @@ export default function GenericClientScreen({
...
+            );
+        }
+
+        return (
+            <>
+                {checkbox}
+                {options}
+            </>
+        );
+    };
```
shazm, line 398: Also something worth thinking about later is potentially moving these snippets into their own little components outside of the APIClient / GenericClient files so that we can reuse them.

src/ApiClient/index.tsx
```diff
@@ -4,17 +4,20 @@
 import { NativeModules } from "react-native";
 import isURL from "validator/es/lib/isURL";
 
-const { APIClient: NativeAPIClient } = NativeModules;
+const { RNNCAPIClient } = NativeModules;
```
shazm, line 7: Are we changing the names from `APIClient` to `RNNCAPIClient` and `GenericClient` to `RNNCGenericClient`? If so - ignore the top comment!

@types/ReactNative.ts
```diff
@@ -5,5 +5,6 @@ declare module "react-native" {
     export interface NativeModulesStatic {
         APIClient: NativeAPIClient;
         GenericClient: NativeGenericClient;
```
shazm, line 7: Just noticed from the changes further below - these probably need to be updated to `RNNCAPIClient` etc if we're changing the names of the native modules.

shazm (changes requested):
Just a few comments - but also double checking that we're changing from `APIClient` to `RNNCAPIClient`?

@types/ReactNative.ts
migbot: I thought about that too, however, since it's importable from `NativeModules` I wonder if it's best to have the prefix to avoid possible (although not probable?) collisions. Seems that most RN libraries have a prefix.

example/src/screens/APIClientScreen.tsx
migbot: Yes absolutely. I'm thinking this should be part of https://mattermost.atlassian.net/browse/MM-30107?

example/src/screens/GenericClientScreen.tsx
migbot: Yep, also can be done as part of https://mattermost.atlassian.net/browse/MM-30107

@types/ReactNative.ts
enahum: Even though I agree with @migbot fo the Emm and PasteInput libraries we did not add any prefix, truth of the matter is that if you are the first to use the name why would we add a prefix, lets have others who want to use the same name to add the prefix themselves.. thoughts?

enahum: I would change it to `NetworkConstants` though

migbot: Sounds good, I'll remove the prefix where I added it and also change to `NetworkConstants`.

@types/ReactNative.ts
```diff
@@ -5,5 +5,6 @@ declare module "react-native" {
...
+        NetworkConstants: NativeConstants;
```
enahum, line 8: lol, too many PR's with the same code

enahum:
got lost, same files being touched, some changes on files that were remove in other PR's... maybe we should bump the speed of reviews instead of basing PR on another PR unless you reference the commits to review, thoughts ?

ios/NetworkConstants-Bridging-Header.h
```diff
@@ -0,0 +1,15 @@
+//
+//  NetworkConstants-Bridging-Header.h
```
migbot, line 2: All these bridging headers are removed in another PR and replaced with one `NetworkClient-Bridging-Header.h`

@types/ReactNative.ts
migbot: Yea, sorry hahaha. #12 was merged so this is good now.

shazm (approved):
All looks good to me; I just see this every so often when trying to build:

```
src/index.tsx:5:38 - error TS1261: Already included file name '/Users/shaz/Code/react-native-network-client/src/APIClient/index.tsx' differs from file name '/Users/shaz/Code/react-native-network-client/src/ApiClient/index.tsx' only in casing.

5 import { getOrCreateAPIClient } from "./APIClient";
```

I think it's from renaming our directory `ApiClient` to `APIClient`.

PR #16, opened 2020-12-04
[MM-31034] iOS support for modifying 3rd party library network requests

migbot:
#### Summary
Swizzling should be the responsibility of the app whose dependency is `react-native-network-client`.

I've added an example of how to swizzle `SDWebImageDownloaderOperation`'s `initWithRequest`method within our example app in order to:
1) Override the session with a new session created from a `react-native-network-client` session configuration. This will ensure that any cookies received from the host server will be sent when making requests with `react-native-fast-image`.
2) Mutate the request to add the authorization header, if any, using our `BearerAuthenticationAdapter`'s `addAuthorizationBearerToken` method.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31034


example/ios/SDWebImageDownloaderOperation+initWithRequest.m
```diff
@@ -0,0 +1,54 @@
...
+  SessionManager *nativeClientSessionManager = [SessionManager default];
+  NSURLSessionConfiguration *configuration = [nativeClientSessionManager getSessionConfigurationFor:host];
+  // If we have a session configured for this host then use this configuration
+  // to create a new session that SDWebImageDownloaderOperation will use for
+  // the request. In addition, if we have an authorization header being added
+  // to our session's requests, then we modify the request here as well using
+  // our BearerAuthenticationAdapter.
+  if (configuration != nil) {
+    NSURLSession *newSession = [NSURLSession sessionWithConfiguration:configuration
+                                                        delegate:session.delegate
```
enahum, line 44: When mutualTlS gets added I guess this delegate needs to be replaced here

ios/APIClient.swift
```diff
@@ -113,17 +113,22 @@ class APIClient: NetworkClient {
...
-            rejectInvalidSession(for: baseUrl, withRejecter: reject)
+        guard let session = SessionManager.default.getSession(for: host) else {
+            rejectInvalidSession(for: host, withRejecter: reject)
             return
         }
 
         let url = baseUrl.appendingPathComponent(endpoint)
+        if let storage = session.session.configuration.httpCookieStorage, let cookies = storage.cookies {
+            for cookie in cookies {
+                storage.deleteCookie(cookie)
```
enahum, line 129: Why delete the cookies?

enahum (commented):
One question

example/ios/SDWebImageDownloaderOperation+initWithRequest.m
migbot: I'll cross that bridge when I get there :)

ios/APIClient.swift
migbot: Haha, woops! I did this to test that just using the bearer auth worked. Going to remove it.

enahum (approved):

PR #17, opened 2020-12-07
Generic Client for Android

shazm:
From our conversation earlier - just splitting up these PR's for ease of review; this one only introduces the GenericClient for Android.


android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -0,0 +1,93 @@
...
+
+        val map = Arguments.createMap()
+        map.putMap("headers", headers)
+        map.putString("data", response.body!!.string())
+        map.putInt("code", response.code)
+        return map;
+    }
+
+    @ReactMethod
+    fun get(baseUrl: String, options: ReadableMap, promise: Promise){
```
migbot, line 30: Since for the GenericClient we are just passing in the full URL.
```suggestion
    fun get(url: String, options: ReadableMap, promise: Promise){
```

shazm: Ah yup! Will update 👍

android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -0,0 +1,93 @@
...
+        val request = Request.Builder().url(baseUrl).build();
+        client.newCall(request).execute().use{ response ->
+            if (response.isSuccessful) {
+                val res = parseResponse(response);
+                return promise.resolve(res)
+            }
+            promise.reject("Unexpected code $response")
```
enahum, line 37: I think you  should reject the promise but include the response even if its not successful so it bubbles up all the way to JS. Same thing with all the other methods

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -0,0 +1,98 @@
...
+                promise.resolve(res)
+            } else {
+                promise.reject(response.code.toString(), res)
+            }
+        }
+    }
+
+    @ReactMethod
+    fun post(url: String, options: ReadableMap, promise: Promise){
+        val body = options.toString().toRequestBody();
```
migbot, line 44: These CRUD methods get [RequestOptions](https://github.com/mattermost/react-native-network-client/blob/1d09b67f572b54aa1969864391b5574614d7bebc/%40types/APIClients.ts#L6) so we should be parsing each option there, no? Or is this done in another PR?

shazm: Yup - I start handling them here; https://github.com/mattermost/react-native-network-client/pull/19

migbot (approved):

enahum (approved):

PR #18, opened 2020-12-07
Network Client for Android

shazm:
From our conversation earlier - just splitting up these PR's for ease of review; this one only introduces the NetworkClient for Android.

Note: Please merge generic client before merging this in.


android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -0,0 +1,127 @@
...
+        val headers = headers.toHashMap()
+        for ((k,v) in headers){
+            sessions[baseUrl]!!.addHeader(k,v as String);
+        }
+        promise.resolve("")
+    }
+
+    @ReactMethod
+    fun invalidateClientFor(baseUrl: String, promise: Promise){
+        sessions.remove(baseUrl);
```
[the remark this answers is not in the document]
shazm: Not on this PR - I think the one's further ahead where I make `sessionClient` and `sessionRequest` separately - I'll just have to clear both.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    var client = OkHttpClient();
+    var sessions = mutableMapOf<String, Request.Builder>()
+
+    override fun getName(): String {
+        return "APIClient"
+    }
+
+    @ReactMethod
+    fun createClientFor(baseUrl: String, config: ReadableMap, promise: Promise){
```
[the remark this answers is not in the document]
enahum: and why is returning an empty string?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    @ReactMethod
+    fun addClientHeadersFor(baseUrl: String, headers: ReadableMap, promise: Promise){
+        val headers = headers.toHashMap()
+        for ((k,v) in headers){
+            sessions[baseUrl]!!.addHeader(k,v as String);
+        }
+        promise.resolve("")
```
enahum, line 44: same here

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    @ReactMethod
+    fun get(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise){
+        val request = sessions[baseUrl]!!.url("$baseUrl/$endpoint").build();
+        client.newCall(request).execute().use{ response ->
+            if (response.isSuccessful) {
+                val res = parseResponse(response);
+                return promise.resolve(res)
+            }
+            promise.reject("Unexpected code $response")
```
enahum, line 72: same comment as in the genericClient

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Updated

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Updated.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -0,0 +1,102 @@
...
+            promise.reject(error)
+        }
+    }
+
+    @ReactMethod
+    fun addClientHeadersFor(baseUrl: String, headers: ReadableMap, promise: Promise){
+        for ((k,v) in headers.toHashMap()){
+            sessions[baseUrl]!!.addHeader(k,v as String);
+        }
+        val headersMap = Arguments.createMap()
```
migbot, line 44: Do we need to return headers here? If so, you could just return `getClientHeadersFor(baseUrl, promise)`?

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    val map = Arguments.createMap()
+    map.putMap("headers", headers)
+    map.putString("data", this.body!!.string())
+    map.putInt("code", this.code)
+    return map;
+}
+
+fun Response.returnPromise(promise: Promise){
+    return if(this.isSuccessful){
```
migbot, line 20: Cam an unsuccessful response have a valid response code? If so, I'd say we always want to resolve the promise so the client doesn't need to catch based on codes.

enahum:
@shazm please merge master and resolve conflicts

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: `promise.resolve()` does need a value, and while we are expecting `Promise<void>` on the App side - I had some feedback around returning the  empty `promise.resolve("")` so I'd rather return the headers for now; I think it may help on the App side as well to ensure the headers are set rather than setting and double checking.

Happy eitherway though :)

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: I believe `response.isSuccessful` only looks at `2XX` codes, anything else is not success. The try/catch is more for IOException that might be thrown by okhttp3, so any unsucessful network calls will be sent back by `promise.reject()`, and any actual errors will be in the try/catch.

PR #19, opened 2020-12-07
 [MM-30121] Android - Follows Redirect

shazm:
PR split - introduces follow-redirect config option on the client.

Note: This requires merging in `android/network-client`.

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -8,7 +8,7 @@ import com.facebook.react.bridge.Arguments;
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
-    var client = OkHttpClient();
+    var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>();
     var sessions = mutableMapOf<String, Request.Builder>()
 
     override fun getName(): String {
@@ -18,7 +18,12 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     @ReactMethod
     fun createClientFor(baseUrl: String, config: ReadableMap, promise: Promise){
         try {
+            sessionsClient[baseUrl] = OkHttpClient().newBuilder();
             sessions[baseUrl] = Request.Builder().url(baseUrl);
+
+            val followRedirect = config.getBoolean("followRedirect")
+            if(followRedirect) sessionsClient[baseUrl]!!.followRedirects(followRedirect)
+
             promise.resolve(sessions[baseUrl]);
         } catch (err: Throwable){
             promise.reject(err)
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -26,7 +26,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
             sessionsClient[baseUrl]?.parseOptions(options);
 
             // Return stringified client for success
-            promise.resolve(sessions[baseUrl].toString())
+            promise.resolve(Unit)
         } catch (err: Throwable) {
             promise.reject(err)
         }
@@ -45,7 +45,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     fun addClientHeadersFor(baseUrl: String, headers: ReadableMap, promise: Promise) {
         try {
             sessions[baseUrl]?.addReadableMap(headers)
-            this.getClientHeadersFor(baseUrl, promise)
+            promise.resolve(Unit);
         } catch (error: Error) {
             promise.reject(error)
         }
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -4,7 +4,7 @@ import com.facebook.react.bridge.*
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
-import com.facebook.react.bridge.Arguments;
+import okio.IOException
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -20,7 +20,12 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
 
     @ReactMethod
     fun createClientFor(baseUrl: String, config: ReadableMap, promise: Promise){
+        sessionsClient[baseUrl] = OkHttpClient().newBuilder();
         sessions[baseUrl] = Request.Builder().url(baseUrl);
+
+        val followRedirect = config.getBoolean("followRedirect")
```
migbot, line 26: What if `followRedirect` is not set on the config?
Also:
```suggestion
        val followRedirects = config.getBoolean("followRedirects")
```

shazm: It just returns a falsy value and doesn't get set; but I believe it's `true` by default so perhaps the false case should be handled.

migbot: Yea that's the same for iOS, so I only [pass it along](https://github.com/mattermost/react-native-network-client/pull/12/files#diff-4c97d446fcd42d8e4eed191e537d4242020da8b5d3ea13a39f3d9f212277ca62R127-R128) if it exists.

enahum (approved):
LGTM other than @migbot comments

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -16,10 +25,65 @@ fun Response.parse (): WritableMap {
...
+ * @params options ReadableMap of options from the App
+ */
+fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builder {
+    this.followRedirects(false);
+    this.followSslRedirects(false);
+
+    val followRedirect = options.getBoolean("followRedirects")
+    if (followRedirect != null) {
+        this.followRedirects(followRedirect)
+        this.followSslRedirects(followRedirect)
```
enahum, line 63:
```suggestion
    val followRedirects = options.getBoolean("followRedirects")
    if (followRedirects != null) {
        this.followRedirects(followRedirects)
        this.followSslRedirects(followRedirects)
```

example/src/screens/APIClientScreen.tsx
```diff
@@ -54,7 +54,7 @@ const styles = StyleSheet.create({
         ...Platform.select({
             ios: { borderColor: PlatformColor("link") },
             android: {
-                borderColor: PlatformColor("?attr/colorControlNormal"),
+                // borderColor: PlatformColor("?attr/colorControlNormal"),
```
enahum, line 57: Why the comment?

src/ApiClient/index.tsx
```diff
@@ -4,7 +4,7 @@
 import { NativeModules } from "react-native";
 import isURL from "validator/es/lib/isURL";
 
-const { RNNCAPIClient } = NativeModules;
+const { APIClient: RNNCAPIClient } = NativeModules;
```
enahum, line 7: Omg, I've seen this changing over and over again. Can you and @migbot come to a decision and stick by it??

I would love to have this resolved asap

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -4,99 +4,124 @@ import com.facebook.react.bridge.*
...
+            sessionsClient[baseUrl] = OkHttpClient().newBuilder();
             sessions[baseUrl] = Request.Builder().url(baseUrl);
-            promise.resolve(sessions[baseUrl]);
-        } catch (err: Throwable){
+
+            // Attach client options
+            sessionsClient[baseUrl]?.parseOptions(options);
+
+            // Return stringified client for success
+            promise.resolve(sessions[baseUrl].toString())
```
enahum, line 29: Not following what sort of string is sent back when the promise gets resolved, can you clarify

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Just the name of the class - but I realise I can pass back kotlin's `Unit` which is essentially `void` in java (`promise.resolve(val)` always expected a `val` so I had to set something...).

example/src/screens/APIClientScreen.tsx
shazm: `borderColor` is still broken on Android :(

shazm:
I've rebased off the latest master and addressed PR feedback, ready for another review

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -4,99 +4,124 @@ import com.facebook.react.bridge.*
...
     @ReactMethod
-    fun addClientHeadersFor(baseUrl: String, headers: ReadableMap, promise: Promise){
-        for ((k,v) in headers.toHashMap()){
-            sessions[baseUrl]!!.addHeader(k,v as String);
+    fun addClientHeadersFor(baseUrl: String, headers: ReadableMap, promise: Promise) {
+        try {
+            sessions[baseUrl]?.addReadableMap(headers)
+            promise.resolve(Unit);
+        } catch (error: Error) {
+            promise.reject(error)
```
migbot, line 50: Let's return explicit errors when possible, like when a [session does not exist](https://github.com/mattermost/react-native-network-client/blob/master/ios/APIClient.swift#L173)

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -4,99 +4,124 @@ import com.facebook.react.bridge.*
...
+            sessionsClient[baseUrl] = OkHttpClient().newBuilder();
             sessions[baseUrl] = Request.Builder().url(baseUrl);
-            promise.resolve(sessions[baseUrl]);
-        } catch (err: Throwable){
+
+            // Attach client options
+            sessionsClient[baseUrl]?.parseOptions(options);
+
+            // Return stringified client for success
+            promise.resolve(Unit)
```
migbot, line 29: What's Unit?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -4,99 +4,124 @@ import com.facebook.react.bridge.*
...
 
     @ReactMethod
-    fun delete(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise){
-        val body = options.toString().toRequestBody();
-        val request = sessions[baseUrl]!!.url("$baseUrl/$endpoint").delete(body).build();
-        client.newCall(request).execute().use{ response ->
-            response.handleViaPromise(promise)
+    fun delete(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise) {
+        try {
+            val body = options.getMap("body").toString().toRequestBody();
```
migbot, line 118: Will this throw an error if options does not include `body`?

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -17,10 +26,65 @@ fun Response.returnAsWriteableMap (): WritableMap {
...
+    return this
+}
+
+/**
+ * Parses options passed in over the bridge for client sessions
+ *
+ * @params options ReadableMap of options from the App
+ */
+fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builder {
+    this.followRedirects(false);
```
migbot, line 58: Alamofire defaults to `true` so can we do the same for Android?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Equivalent to java's `void`, see https://github.com/mattermost/react-native-network-client/pull/19#discussion_r553019456

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: No - it just passes the null object along from my testing.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: I think we should probably make a set of error cases that are handled the same way on both platforms? Happy to write up an error interface in our `@types` that the JS side can expect and we can implement on the native side; though I think it might be bigger task for outside this PR...

@enahum / @avinashlng1080 - any thoughts?

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: Updated 👍

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -4,99 +4,124 @@ import com.facebook.react.bridge.*
...
     override fun getName(): String {
         return "APIClient"
     }
 
     @ReactMethod
-    fun createClientFor(baseUrl: String, config: ReadableMap, promise: Promise){
+    fun createClientFor(baseUrl: String, options: ReadableMap, promise: Promise) {
         try {
+            // Create the client and request builder
+            sessionsClient[baseUrl] = OkHttpClient().newBuilder();
```
enahum, line 22: don't we need to check if the session already exists? or are we ok to override it?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
enahum: I think we should define a map so that we have consistent errors for both platforms instead of returning the raw error. I would expect it to be raw just for the responses

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -17,10 +26,65 @@ fun Response.returnAsWriteableMap (): WritableMap {
...
+    return this
+}
+
+/**
+ * Parses options passed in over the bridge for individual requests
+ *
+ * @param options ReadableMap of options from the App
+ */
+fun Request.Builder.parseOptions(options: ReadableMap): Request.Builder {
+    return this
```
enahum, line 49: todo ?

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -17,10 +26,65 @@ fun Response.returnAsWriteableMap (): WritableMap {
...
+ * Parses options passed in over the bridge for client sessions
+ *
+ * @params options ReadableMap of options from the App
+ */
+fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builder {
+    this.followRedirects(true);
+    this.followSslRedirects(true);
+
+    val followRedirects = options.getBoolean("followRedirects")
+    if (followRedirects) {
```
enahum, line 62: would this evaluate if the value is `false` ? seems that is only being  set when the option is set to `true`

android/src/main/java/com/mattermost/networkclient/NetworkClientPackage.kt
```diff
@@ -7,7 +7,7 @@ import com.facebook.react.uimanager.ViewManager
 
 class NetworkClientPackage : ReactPackage {
     override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> {
-        return listOf<NativeModule>(GenericClientModule(reactContext))
+        return listOf<NativeModule>(GenericClientModule(reactContext), APIClientModule(reactContext), NetworkConstants(reactContext))
```
enahum, line 10: what is `NetworkConstants`?

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
avinashlng1080: Yes a map will be super helpful.  I'm thinking of something like the following can be standardized across both platforms and it may also become useful in the future when we'll implement the 'logger feature'.

```
{
  errorCode : 'xxxxx', // The error code  - e.g. Network_Header_xxxx
  errorClassMethod: 'xxxx',  // APIClientModule - addClientHeadersFor
  errorDescription: 'xxxx xxxxxx xxxxx', // A text description of the error
  errorObject: { xxxx } // The full error object as generated.
}
```

migbot: Created https://mattermost.atlassian.net/browse/MM-31773. @shazm Let's chat/work on this next Monday during our 1:1?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: I think we're okay to override it - on the JS side; when `createClient` is called, I think the assumption is the client doesn't exist already; or maybe you want to create one again - even if there is one on the native side.

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: Almost! It's used in https://github.com/mattermost/react-native-network-client/pull/19 - just rebase conflicts led me to adding this here / now, so I can use it a few PR's later.

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: Oof good spot; this is turning into a catch-22; on the android side `options.getBoolean()` is a non-null, so if it doesn't exist it will return false ergo that if statement doesn't get called. Originally I had it as `if(followRedirects != null)` and that was always true...

@migbot I think we'll have to manually set `followRedirects = true` on the JS side as a "default" when initialising a client to get around this.

android/src/main/java/com/mattermost/networkclient/NetworkClientPackage.kt
shazm: The native network constants we pass back to the JS; things like `EXPONENTIAL_RETRY`.

android/src/main/java/com/mattermost/networkclient/Helper.kt
migbot: @shazm Is there `exists()`-like method on options?

shazm: I found a `options.hasKey()` which is the same idea; using that instead (thanks!).

enahum (approved):

PR #20, opened 2020-12-07
[MM-30124] Android retry connection / connection timeout

shazm:
#### Summary
Introduces connection retries for Android

#### Ticket Link

Jira: https://mattermost.atlassian.net/browse/MM-30124

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/Helper.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Helper.kt
@@ -8,6 +8,7 @@ import okhttp3.Headers
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
+import java.util.concurrent.TimeUnit
 
 /**
  * Parses the response data into the format expected by the App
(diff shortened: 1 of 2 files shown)


@types/APIClients.ts
```diff
@@ -57,6 +57,7 @@ type AndroidAPIClientConfiguration = {
     followRedirects?: boolean;
     timeoutIntervalForRequest?: number;
     timeoutIntervalForResource?: number;
+    retryRequests?: boolean;
```
migbot, line 60: How many retries are attempted and at what intervals? On iOS we have the separate retry configuration:
```
type RetryPolicyConfiguration = {
    type?: NativeConstants["EXPONENTIAL_RETRY"];
    retryLimit?: number;
};
```

shazm: I'm actually not sure - the docs don't specify; I'll go digging in the source - but OkHttp3 briefly mentions rolling your own here: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/retry-on-connection-failure/

shazm: Max is 20; we can build our own interceptor to follow the iOS method using  https://github.com/RunningTheSnail/Okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java

migbot: Here's one from [google-http-java-client](https://github.com/googleapis/google-http-java-client/blob/da1aa993e90285ec18579f1553339b00e19b3ab5/google-http-client/src/main/java/com/google/api/client/util/ExponentialBackOff.java) that might be useful for inspiration.

shazm:
I've added a retry interceptor - it's still a bit buggy / still testing but pushed to show WIP.

I used the [Alamofire](https://github.com/Alamofire/Alamofire/blob/master/Source/RetryPolicy.swift#L313) source for calculating so it's consistent across iOS and Android, and if no retryConfig is passed, the interceptor will not be added to the client (so by default there are no retries).

enahum:
Let me know when this is ready for review

shazm:
Has been cleaned up and corrected, but await https://github.com/mattermost/react-native-network-client/pull/20 to be merged in first.

migbot (approved):
LGTM!

PR #21, opened 2020-12-10
[MM-30117] Per-request timeout config

shazm:
#### Summary

Rather than shallow cloning the clients and deleting, I changed over to using an interceptor that attaches and removes itself from the client for that given request, and sets the timeout for that request in seconds.

#### Ticket Link

Fixes https://mattermost.atlassian.net/secure/RapidBoard.jspa?rapidView=81&modal=detail&selectedIssue=MM-30117


android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -101,15 +101,22 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
         if(timeoutInterval > 0){
-            val duration = timeoutInterval.toLong()
-            sessionsClient[baseUrl]!!.readTimeout(duration, TimeUnit.SECONDS)
+            sessionsClient[baseUrl]!!.addNetworkInterceptor(TimeoutInterceptor(timeoutInterval))
         }
 
         return options;
     }
 
+    private fun parseOptionsClean(baseUrl: String, options: ReadableMap) {
```
migbot, line 110: Could we rename this function to something like `cleanSessionFor` since it's modifying the session client? Likewise for `parseOptions` since it's doing the same and its return value is not used in the caller.

migbot: Or maybe rather than modifying the session in these methods we can just return something that will modify the session in the respective CRUD method? So like:

```
fun get(...) {
   timeoutInterceptor = getTimeoutInterceptorFrom(options)
   if (timeoutInterceptor) {
     sessionsClient[baseUrl]!!.addNetworkInterceptor(TimeoutInterceptor(timeoutInterval))
   }

   sessionsClient[baseUrl]!!.build().newCall(request)...

   cleanSession(session, timeoutInterceptor)
}

fun cleanSession(session, interceptor...) {
  if (interceptor) {
    sessionsClient[baseUrl]!!.networkInterceptors().remove(interceptor)
  }
}
```
Thoughts?

shazm: Yup I think renaming it is probably better; I'm hesitant to modify the session in the methods because we then need to add each config option to each method - that's a lot of duplicated code across the get/post/put/patch/delete - and will be a bit of a pain to update; especially if we add/remote/update a given config option.

I think in psuedo-code it's the difference between:

```
fun get(...){
  modifySessionOnOptions(options);
  ... execute request;
  cleanSessionAsNeeded(options);
}
fun post(...){
  modifySessionOnOptions(options);
  ... execute request;
  cleanSessionAsNeeded(options);
}
...
```

vs

```
fun get(...){
  if(checkForOptionA(options)) doOptionA();
  if(checkForOptionB(options)) doOptionB();
  ... execute request;
   if(checkForOptionA(options)) cleanOptionA();
  if(checkForOptionB(options)) cleanOptionB();
}
fun post(...){
  if(checkForOptionA(options)) doOptionA();
  if(checkForOptionB(options)) doOptionB();
  ... execute request;
   if(checkForOptionA(options)) cleanOptionA();
   if(checkForOptionB(options)) cleanOptionB();
}
```

Pretty sure the potential config options we'll be passing to each method (and even the session initialiser) will be quite a lot eventually.

enahum:
is this ready for review?

shazm:
I've removed the parse/clean methods; we can actually just update the "chain" that makes the call in an interceptor and it only impacts that one call - makes the code much cleaner.

This needs https://github.com/mattermost/react-native-network-client/pull/20 to be merged first.

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -46,7 +46,11 @@ fun Response.promiseResolution(promise: Promise): Response {
  *
  * @param options ReadableMap of options from the App
  */
-fun Request.Builder.parseOptions(options: ReadableMap): Request.Builder {
+fun Request.Builder.parseOptions(options: ReadableMap, session: OkHttpClient.Builder): Request.Builder {
+    val timeoutInterval = options.getInt("timeoutInterval");
+    if(timeoutInterval > 0){
+        session.addNetworkInterceptor(TimeoutInterceptor(timeoutInterval))
```
migbot, line 52: Does the interceptor added to the session affect all future requests?

android/src/main/java/com/mattermost/networkclient/TimeoutInterceptor.kt
```diff
@@ -0,0 +1,25 @@
+package com.mattermost.networkclient
+
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+class TimeoutInterceptor(requestTimeout: Int?) : Interceptor {
```
migbot, line 8: This needs `readTimeout: Int?` as well, no?

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: Yes it does - there wasn't really a way around this; but the Timeout interceptor itself just makes the call on a new "chain" where it attaches the new timeout settings - they don't get carried through.

You do make a valid point; I think a better pattern would be to always have that timeout interceptor attached with a "default" timeout; and if that's ever different - just make the call with the new config on the new / temp "chain".

android/src/main/java/com/mattermost/networkclient/TimeoutInterceptor.kt
shazm: `RequestOptions` only has `timeoutInterval?: number;`; do we want to add `readTimeout` here too? There's a few options in okhttp3 here though; `callTimeout`, `connectTimeout`, `readTimeout` and `writeTimeout`.

android/src/main/java/com/mattermost/networkclient/Helper.kt
migbot: OK, I think I need a better understanding of how these new chains work. Let's jump on a call later today? We can discuss the timeout below as well.

android/src/main/java/com/mattermost/networkclient/TimeoutInterceptor.kt
shazm: Resolved - we're going with request/resource which maps to call/connect/read on android.

android/src/main/java/com/mattermost/networkclient/Helper.kt
shazm: Cleaned up after pair programming session.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -8,8 +8,9 @@ import okio.IOException
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
-    var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>();
-    var sessions = mutableMapOf<String, Request.Builder>()
+    var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>()
+    var sessionsRequest = mutableMapOf<String, Request.Builder>()
+//    var sessionsOptions = mutableMapOf<String, MutableMap<String, Any>>()
```
migbot, line 13:
```suggestion
```

android/src/main/java/com/mattermost/networkclient/HeadersInterceptor.kt
```diff
@@ -0,0 +1,27 @@
...
+class HeadersInterceptor(headers: ReadableMap) : Interceptor {
+
+    private var headers: ReadableMap = headers;
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request();
+        val newRequest = request.newBuilder();
+
+        // Make a new chain; time out should be per-request
```
migbot, line 17:
```suggestion
```

migbot (approved):
Just a couple of comments to remove.

PR #24, opened 2020-12-18
Introduces base "APIClient" for Android

shazm:
A basic `APIClient` is introduced for Android. Further PR's will allow the handling of various configuration options per client / request.

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/Helper.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Helper.kt
@@ -1,0 +1,25 @@
+package com.mattermost.networkclient
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.WritableMap
+import okhttp3.Response
+
+fun Response.returnAsWriteableMap (): WritableMap {
+    val headers = Arguments.createMap();
+    this.headers.forEach{ k -> headers.putString(k.first, k.second) }
+
+    val map = Arguments.createMap()
+    map.putMap("headers", headers)
+    map.putString("data", this.body!!.string())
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -0,0 +1,25 @@
...
+    map.putString("data", this.body!!.string())
+    map.putInt("code", this.code)
+    return map;
+}
+
+fun Response.handleViaPromise(promise: Promise){
+    return if(this.isSuccessful){
+        promise.resolve(this.returnAsWriteableMap());
+    } else {
+        promise.reject(this.code.toString(), this.returnAsWriteableMap())
```
migbot, line 23: If we reject on any non 200-300 responses, then the client would be forced to handle those responses in the `catch` block, no?

enahum: Which is exactly what should happen and what we do today

migbot: Hmm, `fetch` handles a 404 in the `then` block

migbot: Yep, we handle a non-`response.ok` after the fetch call: https://github.com/mattermost/mattermost-mobile/blob/master/app/init/fetch.js#L128-L151

shazm: Added here: https://github.com/mattermost/react-native-network-client/pull/24/files#diff-2fcca6e78100ddac7bdb0e06fdf3d2a6fa5398d7e34ba7415a286d7ed7dd502fR18

migbot (approved):

PR #25, opened 2020-12-18
Android: Native Constants 

shazm:
This adds the `EXPONENTIAL_RETRY` constant for android, and should allow the example to build.

### DIFF
--- a/src/Constants/index.ts
+++ b/src/Constants/index.ts
@@ -4,6 +4,6 @@
 import { NativeModules } from "react-native";
 const { NetworkConstants } = NativeModules;
 
-export default {
-    EXPONENTIAL_RETRY: NetworkConstants.EXPONENTIAL_RETRY,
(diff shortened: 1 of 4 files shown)


src/Constants/index.ts
```diff
@@ -4,6 +4,6 @@
 import { NativeModules } from "react-native";
 const { NetworkConstants } = NativeModules;
 
-export default {
-    EXPONENTIAL_RETRY: NetworkConstants.EXPONENTIAL_RETRY,
-} as NativeConstants;
+const Constants = NetworkConstants.getConstants();
```
migbot, line 7: This won't work for iOS. Maybe:
```
const Constants = Platform.OS === "ios" ? {...NetworkConstants} : NetworkConstants.getConstants();
```
?

migbot (approved):

PR #23, opened 2020-12-18
[MM-31142] JS API for uploads

migbot:
#### Summary
This adds the JS API for uploads and also cleans up the example app to prep for testing uploads once the native implementations are done.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31142

PR #26, opened 2020-12-21
Revert back to creating sessions using baseUrl instead of host

migbot:
#### Summary
This is based on https://github.com/mattermost/react-native-network-client/pull/23. Changes for this PR are in commit https://github.com/mattermost/react-native-network-client/commit/518f4a395d9eda0dce18b7be4ffc5ba63b21258c.

When swizzling SDWebImageDownloader for RN Fast Image, it was easy to find our own session for the image request if we created sessions by using the `host` value rather than the supplied `baseUrl`, since we can easily get the `host` from `request.host` in SDWebImageDownloader. This, however, introduces some other problems (see this [mattermost conversation](https://community.mattermost.com/core/pl/1r89uyif8tf6m8rb5ogrqmqtyy)). This PR reverts back to using the baseUrl for session creation and updates the swizzled SDWebImageDowloader to use a new SessionManager function that attempts to find an existing session by stripping path components from the request URL.

### DIFF
--- a/example/ios/SDWebImageDownloaderOperation+initWithRequest.m
+++ b/example/ios/SDWebImageDownloaderOperation+initWithRequest.m
@@ -31,19 +31,19 @@
 #pragma mark - Method Swizzling
 
 - (nonnull instancetype)swizzle_initWithRequest:(NSURLRequest *)request inSession:(NSURLSession *)session options:(SDWebImageDownloaderOptions)options context:(nullable SDWebImageContext *)context {
-  NSString *host = [[request URL] host];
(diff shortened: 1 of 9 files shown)


enahum (approved):

shazm:
> @shazm Why was the base branch changed?

Ah I noticed in the summary you mentioned it's based off https://github.com/mattermost/react-native-network-client/pull/23 - it just makes it easier using github's diff; once https://github.com/mattermost/react-native-network-client/pull/23 is merged in it's the diff against master.

PR #27, opened 2021-01-09
[MM 31143] iOS upload implementation

migbot:
#### Summary
Based on https://github.com/mattermost/react-native-network-client/pull/23. Changes to be reviewed in this PR are in commits https://github.com/mattermost/react-native-network-client/pull/27/commits/8b61aeb832de38a123d5a70f84c7120933e64437, https://github.com/mattermost/react-native-network-client/pull/27/commits/917e5c66f29760ea51ef9f6f0ee362972662b94a, and https://github.com/mattermost/react-native-network-client/pull/27/commits/c5639d14c7c2d4ec481af87dfe4708cb47ff7e2f

* Uses Alamofire's `session.upload` for uploading from a file URL. If `skipBytes` is passed, we seek to the byte offset of the file, read to the end of the file from that offset, and upload that data.
* `NativeClient-UploadProgress` event is emitted to [track progress](https://github.com/mattermost/react-native-network-client/pull/27/commits/8b61aeb832de38a123d5a70f84c7120933e64437#diff-89f404459b27cb08479a7fedf192808a92b142baa24474f4fbd2b1e2e122ba3dR85-R120)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31143

PR #28, opened 2021-01-11
[MM 31143] iOS upload implementation

migbot:
#### Summary
* Uses Alamofire's `session.upload` with an `InputStream` for uploading from a file URL. If `skipBytes` is passed its value is used to set the `fileCurrentOffsetKey` of the stream.
* `NativeClient-UploadProgress` event is emitted to [track progress](https://github.com/mattermost/react-native-network-client/pull/27/commits/8b61aeb832de38a123d5a70f84c7120933e64437#diff-89f404459b27cb08479a7fedf192808a92b142baa24474f4fbd2b1e2e122ba3dR85-R120)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31143


@types/NativeAPIClient.ts
```diff
@@ -1,53 +1,61 @@
...
-        endpoint: string | null,
-        options?: RequestOptions
-    ): Promise<ClientResponse>;
-    upload(
-        baseUrl: string,
-        endpoint: string | null,
-        fileUrl: string,
-        options?: UploadRequestOptions
-    ): Promise<ClientResponse>;
+import type { NativeModule } from "react-native"; // eslint-disable-line no-unused-vars
```
enahum, line 4: why the eslint comment?

example/@types/navigation.ts
```diff
@@ -15,6 +15,7 @@ declare global {
     type APIClientItem = {
         name: string;
         client: APIClientInterface;
+        isMattermostClient?: boolean;
```
enahum, line 18: ????

example/ios/NetworkClientExample/AppDelegate.m
```diff
@@ -11,6 +11,7 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
 
+#if DEBUG
```
enahum, line 14: ???

example/src/utils/index.tsx
```diff
@@ -21,7 +21,7 @@ export const parseHeaders = (headers: Header[]): ClientHeaders => {
 
 export const createMattermostAPIClient = async (): Promise<APIClientItem | null> => {
     const name = "Mattermost";
-    const baseUrl = "https://community.mattermost.com";
+    const baseUrl = "http://community.mattermost.com";
```
enahum, line 24: what happened with ssl?

ios/APIClient.m
```diff
@@ -29,6 +32,6 @@ @interface RCT_EXTERN_MODULE(APIClient, NSObject)
 
 RCT_EXTERN_METHOD(delete:(NSString *)baseUrl forEndpoint:(NSString *)endpoint withOptions:(NSDictionary *)options withResolver:(RCTPromiseResolveBlock)resolve withRejecter:(RCTPromiseRejectBlock)reject)
 
-RCT_EXTERN_METHOD(upload:(NSString *)baseUrl forEndpoint:(NSString *)endpoint withFileUrl:(NSString *)fileUrlString withOptions:(NSDictionary *)options withResolver:(RCTPromiseResolveBlock)resolve withRejecter:(RCTPromiseRejectBlock)reject)
+RCT_EXTERN_METHOD(upload:(NSString *)baseUrl forEndpoint:(NSString *)endpoint withFileUrl:(NSString *)fileUrlString withTaskId:(NSString *)taskId withOptions:(NSDictionary *)options withResolver:(RCTPromiseResolveBlock)resolve withRejecter:(RCTPromiseRejectBlock)reject)
```
enahum, line 35: what is the taskid?

enahum (commented):
asking about task id before I complete the review

example/@types/navigation.ts
migbot: There are two upload screens in the example app, one is for a Mattermost upload which handles API requests to create a session, upload and post to a channel, and resume a failed upload. `isMattermostClient` is used to determine whether there should be a `Mattermost Upload` button in the client screen to navigate to that separate Mattermost upload screen.

example/src/utils/index.tsx
migbot: Typo. Fixing.

@types/NativeAPIClient.ts
migbot: I was getting `4:15  error  'NativeModule' is defined but never used  no-unused-vars`, but fixed just now with [this](https://github.com/mattermost/react-native-network-client/pull/28/commits/3f7b49c6b9fde096fd673cbcf073d7f9d67fe77b#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR30)

ios/APIClient.m
migbot: The progress event emitted includes the [`taskId`](https://github.com/mattermost/react-native-network-client/pull/28/files#diff-ceeeb3e5b18acce6df7a9516ad89c5babbf12e2204a2cc41569582fc0678afbeR171) which the JS side then used to determine if it's [`onProgress` callback should be called](https://github.com/mattermost/react-native-network-client/pull/28/files#diff-b7b269eb597141b5d1ba273f02146e50bba7355978f9e5b60a3d63c0408f4191R92-R94)

example/ios/NetworkClientExample/AppDelegate.m
migbot: I was getting an error when building without this. Checking again.

migbot: OK, removed.

enahum: I think you missed this one

ios/APIClient.swift
```diff
@@ -111,6 +126,87 @@ class APIClient: NetworkClient {
...
+        if let skipBytes = options["skipBytes"].uInt64 {
+            do {
+                let data: Data?
+                let fileHandle = try FileHandle.init(forReadingFrom: fileUrl)
+                if #available(iOS 13.4, *) {
+                    try fileHandle.seek(toOffset: skipBytes)
+                    data = try? fileHandle.readToEnd()
+                } else {
+                    fileHandle.seek(toFileOffset: skipBytes)
+                    data = fileHandle.readDataToEndOfFile()
```
enahum, line 164: what happens if the file is large enough, wouldn't this be a problem ?

migbot: Yep, problem. Will fix.

src/APIClient/index.tsx
```diff
@@ -20,6 +21,10 @@ const DEFAULT_API_CLIENT_CONFIG: APIClientConfiguration = {
     },
 };
 
+const generateUploadTaskId = () =>
+    Math.random().toString(36).slice(-10) +
+    Math.random().toString(36).slice(-10);
+
```
shazm, line 27: From what I understand RN is essentially single-threaded so there will only ever be a single upload at a time when it's called, I don't think this will be needed?

src/APIClient/index.tsx
```diff
@@ -77,8 +82,42 @@ class APIClient implements APIClientInterface {
...
+            }
+        );
+
+        promise.progress = (fn) => {
+            promise.onProgress = fn;
+            return promise;
+        };
+
+        return promise;
+    };
```
shazm, line 120: Just following on from the top comment; I think it would simplify this method if we just handled it via a callback? The other option is also to emit the event, but offer a helper that can listen in? Or both?

shazm (commented):
Just a few thoughts as I'm working through the android uploader around simplifying this a little bit.

src/APIClient/index.tsx
enahum: @shazm JS may be single threaded but the native side is not, right?

@types/APIClients.ts
```diff
@@ -1,6 +1,13 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+interface ProgressPromise<T> extends Promise<T> {
+    progress?: (
+        callback: (fractionCompleted: number) => void
+    ) => ProgressPromise<T>;
+    onProgress?: (fractionCompleted: number) => void;
+}
+
```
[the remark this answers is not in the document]
enahum: I believe Miguel is following RNFetchBlob here,  0/5 if we use one or the other, but we should decide on the pattern

@types/NativeAPIClient.ts
```diff
@@ -1,53 +1,61 @@
...
-        options?: UploadRequestOptions
-    ): Promise<ClientResponse>;
+import type { NativeModule } from "react-native";
 
-    createClientFor(
-        baseUrl: string,
-        config?: APIClientConfiguration
-    ): Promise<void>;
+declare global {
+    interface NativeAPIClient extends NativeModule {
```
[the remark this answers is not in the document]
migbot: I had to do this in order to do `new NativeEventEmitter(NativeAPIClient)` as `NativeEventEmitter` requires a `NativeModule` as a param.

@types/APIClients.ts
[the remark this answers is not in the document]
migbot: Yea, just following RNFetchBlob here. I'll take a look at axios.

ios/APIClient.swift
```diff
@@ -111,6 +127,89 @@ class APIClient: NetworkClient {
...
+        let stream = InputStream(url: fileUrl)!
+        if let skipBytes = options["skipBytes"].uInt64 {
+            stream.setProperty(skipBytes, forKey: .fileCurrentOffsetKey)
+            initialFractionCompleted = Double(skipBytes) / fileSize
+        }
+
+        let request = session.upload(stream, to: url, headers: headers, interceptor: interceptor, requestModifier: requestModifer)
+            .uploadProgress { progress in
+                if (self.hasListeners) {
+                    let fractionCompleted = initialFractionCompleted + (Double(progress.completedUnitCount) / fileSize)
```
migbot, line 172: The `progress` object here usually has a `fractionCompleted` property that is calculated from the `completedUnitCount` and the `totalUnitCount`, however, for streams the `totalUnitCount` is always `-1`. For this reason we calculate it here using the `fileSize`.

Also, an `initialFractionCompleted` is added if the upload is being resumed. This allows the progress to pick up from where it left off, instead of resetting to 0.

ios/APIClient.swift
```diff
@@ -111,6 +127,89 @@ class APIClient: NetworkClient {
...
+                            "headers": nil,
+                            "data": nil,
+                            "code": error.responseCode,
+                            "lastRequestedUrl": nil
+                        ])
+
+                        return
+                    }
+
+                    reject("\(error.responseCode)", error.errorDescription, error)
```
migbot, line 200: Errors will be standardized with https://mattermost.atlassian.net/browse/MM-31773

ios/APIClient.swift
```diff
@@ -111,6 +127,89 @@ class APIClient: NetworkClient {
...
+                }
+            }
+
+        self.requestsTable.setObject(request, forKey: taskId as NSString)
+    }
+    
+    @objc(cancelRequest:withResolver:withRejecter:)
+    func cancelRequest(_ taskId: String, withResolver resolve: RCTPromiseResolveBlock, withRejecter reject: RCTPromiseRejectBlock) -> Void {
+        if let request = self.requestsTable.object(forKey: taskId as NSString) {
```
migbot, line 209: I followed the RNFetchBlob way of using an `NSMapTable` for tracking requests and allowing cancellation. I added cancellation just for uploads as another way to test resumes. Request cancellation in general will be added with https://mattermost.atlassian.net/browse/MM-32126

enahum (approved):
nice job with using InputStream

@types/APIClients.ts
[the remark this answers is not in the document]
shazm: After a bit of back and forth, this is probably a better way. 

Passing the callback in via config options seemed like an anti-pattern, and actually had it's own set of complications.

shazm (approved):
All good - I'm tiding up the usage of ProgressPromise with the Android file upload so this is good :)

PR #30, opened 2021-01-16
Android Tidy

shazm:
#### Summary

I was having a few issues with the android emulator / kotlin and some related things; this tidies up the Kotlin code and compiles without needing to mess around with Android Studio / SDK's / Versions 🎉


android/src/main/java/com/mattermost/networkclient/APIClient.kt
```diff
@@ -6,7 +6,7 @@ import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
 import okio.IOException
 
-class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
+class APIClient(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
```
enahum, line 9: why the change in the class name?

android/src/main/java/com/mattermost/networkclient/APIClient.kt
```diff
@@ -26,7 +26,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
             sessionsClient[baseUrl]!!.parseOptions(options);
 
             // Return stringified client for success
-            promise.resolve(Unit)
+            promise.resolve( null)
```
enahum, line 29: I thought `Unit` = void? what happened here?

android/src/main/java/com/mattermost/networkclient/APIClient.kt
```diff
@@ -124,4 +124,14 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+
+    @ReactMethod
+    fun upload(baseUrl: String, endpoint: String?, fileUrl: String, taskId: String, options: ReadableMap, promise: Promise) {
+
+    }
+
+    @ReactMethod
+    fun cancelRequest(taskId: String) {
+
+    }
```
enahum, line 136: are these supposed to be part of this PR?

android/src/main/java/com/mattermost/networkclient/GenericClient.kt
```diff
@@ -4,10 +4,9 @@ import com.facebook.react.bridge.*
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
-import okhttp3.Response
 
 
-class GenericClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
+class GenericClient(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
```
enahum, line 9: same as above

enahum (commented):
There are a few things that I need to understand before approving this

android/src/main/java/com/mattermost/networkclient/APIClient.kt
shazm: Oh happy to leave as `APIClientModule`, it just bugged me a little that we call it `APIClient` whereas `NetworkConstants` is name on both the file / class, and returned via `getName()` so just trying to be more consistent. Happy either way, but not both at the same time 😄

android/src/main/java/com/mattermost/networkclient/APIClient.kt
shazm: I kept having some issues with the Java `Void` class getting initialised here (and attempts to send this class over the bridge in the `promise.resolve`), when I really just wanted the `void` type - but weirdly they both ended up coming across as `null` over the bridge so figured `null` could just do rather than digging too deep into the Kotlin `Unit` >> Java `Void` interop

android/src/main/java/com/mattermost/networkclient/APIClient.kt
shazm: Yep; there's the upload PR incoming shortly - just git rebase conflict avoidance.

android/src/main/java/com/mattermost/networkclient/APIClient.kt
enahum: I get you, but I think we should keep the RN way of doing things here, for consistency. So we have a Package and a Module file and the name exported to JS is without `Module` in the name

shazm: Yup! In that case I'll call it `NetworkConstantsModule` for the android side for consistency 👍

android/src/main/java/com/mattermost/networkclient/APIClient.kt
enahum: If this is not part of the PR we should remove it, I rather keep the history clean than avoiding having to resolve conflicts when we submit PR's on top of each other (which we should also avoid).

android/src/main/java/com/mattermost/networkclient/APIClient.kt
enahum: is `NetworkConstants` a module? or is it just constants? I'm struggling to understand why we have a module named `NetworkConstants`

android/src/main/java/com/mattermost/networkclient/APIClient.kt
shazm: Easy, no probs - I'll remove.

android/src/main/java/com/mattermost/networkclient/APIClient.kt
shazm: Ah so technically it is just another module; it's initialised the same way as `APIClient` or `GenericClient` is - it just holds a `getConstants()` method which returns an object of constants.

With TurboModules coming - it will be exactly like any of the other methods too; https://reactnative.dev/docs/native-modules-android#exporting-constants

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -73,40 +76,42 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builde
...
+        val retryLimit = retryPolicyConfiguration["retryLimit"] as Double?
+        val retryExponentialBackoffBase = retryPolicyConfiguration["exponentialBackoffBase"] as Double?
+        val retryExponentialBackoffScale = retryPolicyConfiguration["exponentialBackoffScale"] as Double?
+        this.addInterceptor(RetryInterceptor(retryType, retryLimit?.toInt(), retryExponentialBackoffBase, retryExponentialBackoffScale))
     }
 
     // Headers
-    if(options.hasKey("headers")){
+    if (options.hasKey("headers")) {
         this.addNetworkInterceptor(HeadersInterceptor(options.getMap("headers")!!));
```
migbot, line 92: What's the difference between `addNetworkInteceptor` and `addInterceptor`?

shazm: `addNetworkInteceptor` can't call proceed more than once, because they act on the intermediately requests/responses; so they kick in before/after things like redirects - whereas `addInterceptor` doesn't care about those, plus a few more things details here:

https://square.github.io/okhttp/interceptors/#choosing-between-application-and-network-interceptors

shazm:
Has been rebased off master and ready for a review / merge.

migbot (approved):

enahum (approved):

PR #37, opened 2021-01-22
[MM-31901] [MM-31902] WebSocket JS API and iOS implementation

migbot:
#### Summary
Uses [Starscream](https://github.com/daltoniam/Starscream) for iOS WebSockets

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31901
https://mattermost.atlassian.net/browse/MM-31902


ios/WebSocketClient.swift
```diff
@@ -0,0 +1,146 @@
...
+        reject("\(error.code)", message, error)
+    }
+    
+    func didReceive(event: WebSocketEvent, client: WebSocket) {
+        let url = client.request.url!.absoluteString
+
+        switch event {
+        case .connected(let headers):
+            self.sendEvent(withName: EVENTS["OPEN_EVENT"], body: ["url": url, "message": ["headers": headers]])
+            self.sendEvent(withName: EVENTS["READY_STATE_EVENT"], body: ["url": url, "message": READY_STATE["OPEN"]!])
```
migbot, line 127: I'm not liking sending two events here. Since we don't use `readyState` on the native side, we could just opt to set that value on the JS side when `OPEN_EVENT` and `CLOSE_EVENT` are received. See https://github.com/mattermost/react-native-network-client/commit/583589cbef06058ba1d45034098b18185d30dfb1. Thoughts?

migbot: Another option is to just add `readyState` to the body of the `OPEN_EVENT` and `CLOSE_EVENT` but I think that just achieves the same result as https://github.com/mattermost/react-native-network-client/commit/583589cbef06058ba1d45034098b18185d30dfb1

@types/NativeWebSocketClient.ts
```diff
@@ -0,0 +1,36 @@
...
+    interface NativeWebSocketClient extends NativeModule {
+        getConstants(): WebSocketConstants;
+        createClientFor(
+            url: string,
+            config?: WebSocketClientConfiguration
+        ): Promise<void>;
+        connectFor(url: string): Promise<void>;
+        disconnectFor(url: string): Promise<void>;
+        sendDataFor(url: string, data: string): Promise<void>;
+    }
```
shazm, line 35: I would add the two missing event related methods here rather than importing and then needing to wrap in a declare global.

@types/NativeWebSocketClient.ts
```diff
@@ -0,0 +1,36 @@
...
+        MESSAGE_EVENT = "NetworkClient-WebSocket-Message",
+        READY_STATE_EVENT = "NetworkClient-WebSocket-ReadyState",
+    }
+
+    enum WebSocketReadyState {
+        CONNECTING,
+        OPEN,
+        CLOSING,
+        CLOSED,
+    }
```
shazm, line 19: Should we have these in utils instead?

@types/WebSocketClient.ts
```diff
@@ -1,19 +1,37 @@
...
-    | AndroidWebSocketClientConfiguration;
+type WebSocketCallback = (event: WebSocketEvent) => void;
+
+interface WebSocketClientInterface {
+    url: string;
+    readyState: WebSocketReadyState;
+    send(data: string): void;
+    connect(): void;
+    close(): void;
+    onOpen(callback: WebSocketCallback): void;
```
shazm, line 33: Just for consistency; we should do either `connect` / `onConnect` or `open` / `onOpen`?

MDN docs uses open; and the constructor just connects - so perhaps we should stick with `open()` and `onOpen()`?

example/package.json
```diff
@@ -23,14 +23,17 @@
...
     "react-native-numeric-input": "1.8.3",
     "react-native-permissions": "3.0.0",
     "react-native-progress": "4.1.2",
     "react-native-reanimated": "1.13.1",
     "react-native-safe-area-context": "3.1.8",
     "react-native-screens": "2.12.0",
-    "react-native-vector-icons": "7.1.0"
+    "react-native-vector-icons": "7.1.0",
+    "react-redux": "^7.2.2",
+    "redux": "^4.0.5"
```
shazm, line 36: Are these being used?

ios/WebSocketClient.swift
shazm: I don't mind having two events; but I do like having `readyState` as an actual prop / state that we can pass around in the JS side; so we don't have to listen to the event and can just use it; i.e. `readyState === 1 ? {showX} : {showY}`

shazm (commented):
Just some feedback items, but otherwise it looks pretty solid. I'll have some more feedback around the SSL specific stuff as I do the equivalent on Android, as I think OkHttp3 handles it slightly differently to iOS.

example/package.json
migbot: Nope, will remove!

@types/WebSocketClient.ts
migbot: Yep, will change to `open`.

@types/NativeWebSocketClient.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    type WebSocketConstants = {
+        EVENTS: typeof WebSocketEvents;
+        READY_STATE: typeof WebSocketReadyState;
+    };
+
+    interface NativeWebSocketClient extends NativeModule {
+        getConstants(): WebSocketConstants;
```
migbot, line 27: @shazm If the enums are in utils, I'd then have to import them to use `WebSocketConstants` here no? And then I'd have to wrap in `declare global`?

@types/NativeWebSocketClient.ts
```diff
@@ -0,0 +1,36 @@
...
+    CONNECTING,
+    OPEN,
+    CLOSING,
+    CLOSED,
+}
+
+type WebSocketConstants = {
+    EVENTS: typeof WebSocketEvents;
+    READY_STATE: typeof WebSocketReadyState;
+};
```
shazm, line 22: 👍

@types/WebSocketClient.ts
```diff
@@ -1,19 +1,37 @@
...
 // See LICENSE.txt for license information.
 
-interface WebSocketClientInterface {
-    wsUrl: string;
-    disconnect(): void;
-}
+type SSLPinningConfiguration = {
+    enabled: boolean;
+    allowSelfSigned?: boolean;
+};
```
shazm, line 7: I think you also need to pass in the certificate and host to pin as well?

example/src/components/ResponseOverlay.tsx
```diff
@@ -18,6 +18,7 @@ const ResponseOverlay = ({
 }: ResponseOverlayProps) => {
     const hide = () => setVisible(false);
 
+    console.log(response?.headers);
```
shazm, line 21: Should remove this?

shazm (approved):
Just a few minor things, but I can also clean them up as part of the android PR :)

@types/WebSocketClient.ts
migbot: Yep yep, I can handle that in the mTLS PR.

example/src/components/ResponseOverlay.tsx
migbot, line 21:
```suggestion
```

PR #38, opened 2021-02-02
[MM-31035] Android TLS + Swizzle

shazm:
This is my first pass at introducing TLS certificates to the OkHttp3 client, and then using a global interceptor that replaces any OkHttp3 requests with our client if the base url's match.

Things to note for swizzling:

- Hunter library is installed in the example app, and calls the GlobalInterceptor from the package
- Native Network package now uses a singleton for making the client session objects and requests available across the package
- Global Interceptor calls this singleton to get client session

^^ This is still a bit finicky at the moment, I may have to use delegates I think

Things to note for tls certificates:

- Certificates are expected as PEM strings from the JS side
- Certificates are only attached to the client session (not globally!)
- Because of the above, the global interceptor then looks at the request before replacing the previous client with our own
- Certificate pinning also only works on a given client session

^^ I haven't tested this yet, currently still working through the swizzling.

I also had to update gradle in the example app so that the hunter library would work

### DIFF
--- a/@types/APIClients.ts
+++ b/@types/APIClients.ts
@@ -58,6 +58,13 @@ interface APIClientInterface {
     invalidate(): Promise<void>;
 }
 
+type CertificateConfiguration = {
+    rootCertificate?: string;
+    pinnedCertificate?: string;
+    pinnedHost?: string;
+    clientCertificate?: string;
+};
+
 type SessionConfiguration = {
     followRedirects?: boolean;
     allowsCellularAccess?: boolean;
@@ -86,6 +93,7 @@ type APIClientConfiguration = {
     requestAdapterConfiguration?: RequestAdapterConfiguration;
     serverTrustManagerConfig?: Record<string, string>;
     cachedResponseHandlerConfig?: Record<string, string>;
+    certificateConfig?: CertificateConfiguration;
 };
 
 type UploadProgressEvent = {
--- a/android/src/main/java/com/mattermost/networkclient/Helper.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Helper.kt
@@ -6,7 +6,7 @@ import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableMap
 
 import okhttp3.*
-import okhttp3.tls.Certificates;
+//import okhttp3.tls.Certificates;
 import okhttp3.tls.HandshakeCertificates;
 import java.util.concurrent.TimeUnit
 
@@ -140,9 +140,11 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builde
 
 
             // Add the client certificate if provided
-            if(certsConfig.hasKey("clientCertificatePem")) {
+            if(certsConfig.hasKey("clientCertificate")) {
                 val clientCertificate = HeldCertificate.decode(certsConfig.getString("clientPublicCertificate")!!)
                 certificatesBuilder = certificatesBuilder.heldCertificate(clientCertificate)
+            } else {
+
             }
 
             // Build and add to the client
--- a/android/src/main/java/com/mattermost/networkclient/Helper.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Helper.kt
@@ -6,9 +6,14 @@ import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableMap
 
 import okhttp3.*
+import okhttp3.tls.Certificates;
+import okhttp3.tls.HandshakeCertificates;
 import java.util.concurrent.TimeUnit
 
 import com.mattermost.networkclient.interceptors.*
+import okhttp3.tls.HeldCertificate
+import okhttp3.tls.decodeCertificatePem
+import java.security.cert.X509Certificate
 
 
 /**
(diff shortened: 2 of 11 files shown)


example/android/build.gradle
```diff
@@ -12,10 +12,13 @@ buildscript {
...
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:3.5.3")
+        classpath('com.android.tools.build:gradle:4.1.2')
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
+
+        classpath 'com.quinn.hunter:hunter-okhttp-plugin:1.1.0'
+        classpath 'com.quinn.hunter:hunter-transform:1.1.0'
```
shazm, line 21: So the RN code says not to put dependencies here, but this fails when I place it in the individual module - any thoughts around this would be great.

@types/APIClients.ts
```diff
@@ -58,6 +58,13 @@ interface APIClientInterface {
     invalidate(): Promise<void>;
 }
 
+type CertificateConfiguration = {
```
avinashlng1080, line 61: This is a style preference.  I normally sort everything alphabetically as this makes it more developer-friendly and it is easy on the eyes for when you need to debug and deep dive into codes.  If you are looking for `clientCertificate`, you need to traverse 3 lines to get to it.

example/android/build.gradle
avinashlng1080: What error message do you get when it failed ?

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -118,6 +123,42 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builde
...
+            var certificates: HandshakeCertificates;
+            var certificatesBuilder = HandshakeCertificates
+                    .Builder()
+                    .addTrustedCertificate(rootCertificate)
+                    .addPlatformTrustedCertificates()
+
+
+            // Add the client certificate if provided
+            if(certsConfig.hasKey("clientCertificate")) {
+                val clientCertificate = HeldCertificate.decode(certsConfig.getString("clientPublicCertificate")!!)
```
migbot, line 144: Why is the client cert decoded with `HeldCertificate` while the root cert with `decodeCertificatePem()`?

android/src/main/java/com/mattermost/networkclient/Helper.kt
```diff
@@ -118,6 +123,42 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap): OkHttpClient.Builde
...
+
+            }
+
+            // Build and add to the client
+            certificates = certificatesBuilder.build()
+            this.sslSocketFactory(certificates.sslSocketFactory(), certificates.trustManager)
+        }
+
+        // If a "pinned" certificate and host are given, pin them!
+        if(certsConfig.hasKey("pinnedHost") && certsConfig.hasKey("pinnedCertificate") ){
```
migbot, line 156: Do we want to return an error if only one key is provided?

android/src/main/java/com/mattermost/networkclient/interceptors/GlobalInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+
+        // Get our singletons
+        val clientSessions = SessionsObject.client
+        var request = chain.request();
+        val interceptUrl = request.url.host
+
+        // No client session found, proceed as is
+        if (clientSessions.isNullOrEmpty() || clientSessions[interceptUrl] !== null) {
```
migbot, line 19:
```suggestion
        if (clientSessions.isNullOrEmpty() || clientSessions[interceptUrl] == null) {
```

migbot: Also, will the `request.url.host` format always match the format of the url `clientSessions`? For example, could the url in `clientSessions` be `https://some-url.com/` (with trailing slash) and the `request.url.host` be `https://some-url.com` (no trailing slash), causing the session not to be found?

android/src/main/java/com/mattermost/networkclient/interceptors/GlobalInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+            return chain.proceed(request)
+        }
+
+        // Set our client session
+        var client = clientSessions[interceptUrl]!!;
+
+        // Execute request
+        return client.build().newCall(request).execute()
```
migbot, line 27: This might not be a valid question but if we're not using `chain` here, what happens to any other interceptors that might have been attached to the request?

android/src/main/java/com/mattermost/networkclient/interceptors/PinnedCertificateInterceptor.kt
```diff
@@ -0,0 +1,22 @@
...
+import java.net.InetAddress
+
+
+class PinnedCertificateInterceptor(private val pinnedCert: String, private val pinnedHost: String) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request();
+        val call = chain.call();
+        val connection = chain.connection()
```
migbot, line 17:
```suggestion
        chain.call();
        chain.connection()
```

shazm:
Closed in favour of https://github.com/mattermost/react-native-network-client/pull/55 - Swizzling will be handled in a seperate PR.

PR #39, opened 2021-02-04
[MM-31903] Android Websockets

shazm:
#### Summary
Introduces websocket support for Android.

#### Ticket Link
Resolves https://mattermost.atlassian.net/browse/MM-31903?atlOrigin=eyJpIjoiODA4NzUyNzNmYzM4NGE2OTk5YzUyNTZjMjM3NTk0ZGYiLCJwIjoiaiJ9

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
@@ -1,0 +1,71 @@
+package com.mattermost.networkclient
+
+import com.facebook.react.bridge.*
+import com.mattermost.networkclient.events.WebSocketEvent
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.WebSocket
+import kotlin.collections.HashMap
+
+class WebSocketClientModule(private val reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
+
+    var clients = mutableMapOf<String, OkHttpClient.Builder>()
+    var sockets = mutableMapOf<String, WebSocket>()
+    var requests = mutableMapOf<String, Request.Builder>()
+
+    override fun getName(): String {
+        return "APIClient"
+    }
+
+    @ReactMethod
+    fun createClientFor(url: String, options: ReadableMap, promise: Promise) {
+        try {
+            // Create the client and request builder
+            clients[url] = OkHttpClient().newBuilder();
+            requests[url] = Request.Builder().url(url);
+
+            // Attach client options if they are passed in
+            clients[url]!!.parseOptions(options);
+
+            // Return stringified client for success
+            promise.resolve(null)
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/mattermost/networkclient/events/WebSocketEvent.kt
```diff
@@ -0,0 +1,41 @@
...
+        emitEvent(event)
+    }
+
+    override fun onFailure(webSocket: WebSocket, t: Throwable, response: Response?) {
+        val event = Arguments.createMap();
+        event.putString("NetworkClient-WebSocket-Error", t.message)
+        emitEvent(event)
+    }
+
+    override fun onClosing(webSocket: WebSocket, code: Int, reason: String) {
```
migbot, line 28: Is this event needed? If so I'll need to update the iOS side to emit it prior to [disconnect](https://github.com/mattermost/react-native-network-client/blob/master/ios/WebSocketClient.swift#L91).

ios/APIClient.swift
```diff
@@ -26,6 +26,14 @@ class APIClient: RCTEventEmitter, NetworkClient {
...
     }
 
+    func requiresMainQueueSetup() -> Bool {
+        return false
+    }
+    
+    override func constantsToExport() -> [AnyHashable : Any]! {
+        return CONSTANTS
+    }
+
```
migbot, line 36:
```suggestion
```
These are already defined.

android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
```diff
@@ -0,0 +1,71 @@
...
+            sockets[url]!!.send(data)
+        } catch (err: Throwable) {
+            promise.reject(err)
+        }
+    }
+
+    @Override
+    override fun getConstants(): Map<String, Any> {
+        val constants: MutableMap<String, Any> = HashMap<String, Any>()
+        constants["EVENTS"] = HashMap<String, String>()
```
migbot, line 67: I don't see the map of events defined. Maybe they can be `public static final` contants in class `WebSocketEvent` and then you can list them here as `WebSocketEvent.OPEN_EVENT`, etc?

android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
```diff
@@ -0,0 +1,71 @@
...
+        constants["READY_STATE"] = HashMap<String, String>()
```
migbot, line 68: I don't see where these are defined either.

android/src/main/java/com/mattermost/networkclient/events/WebSocketEvent.kt
shazm: It's not used but I extend the EventListener on the Android side, I need to have this method to not get errors.

migbot (approved):

PR #43, opened 2021-03-03
iOS - Client certificate authentication

migbot:
#### Summary
Adds client certificate authentication for the `APIClient` and `WebSocketClient`. For both `APIClient` and `WebSocketClient` their configuration object accepts a path to a `.p12` file and an optional password. For just the `APIClient`, an `importClientP12` has also been added to allow for importing of the `.p12` file after the client has been created. When the SSL handshake requires client authentication and the client certificate is not found, an `APIClient-Client-Certificate-Missing` is sent to the JS side. The example app shows how to subscribe and react to this event.

I also changed some of the TS types and native constants.

Notes:
* The error handling will be done as part of https://mattermost.atlassian.net/browse/MM-31773
* The `WebSocketClient` required changes to `Starscream` made in this fork's `client-auth-support` branch: https://github.com/mattermost/Starscream/tree/client-auth-support

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-31981


enahum (approved):
this looks really good

shazm (approved):

example/src/utils/index.tsx
```diff
@@ -71,6 +71,9 @@ const buildDefaultApiClientConfiguration = (
         exponentialBackoffBase: 2,
         exponentialBackoffScale: 0.5,
     };
+    const requestAdapterConfiguration = {
```
josephbaylon, line 74: @migbot This is already declared as parameter in line 55. This was moved to parameter so that other clients can pass a different value.

migbot: Woops, will remove.

example/src/utils/index.tsx
```diff
@@ -71,6 +71,9 @@ const buildDefaultApiClientConfiguration = (
...
+        bearerAuthTokenResponseHeader: "Token",
+    };
```
migbot, line 76:
```suggestion
```

josephbaylon:
@migbot Per our conversation, the app crashes when it attempts to upload

example/src/utils/index.tsx
```diff
@@ -248,12 +248,14 @@ const createMattermostWebSocketClient = async (): Promise<WebSocketClientItem |
     const name = "Mattermost Web Socket";
     const host =
         Platform.OS === "ios"
-            ? "ws://192.168.0.14:8065"
-            : "ws://10.0.2.2:8065";
-    const url = `${host}/api/v4/websocket`;
+            ? "192.168.0.14:8065"
+            : "10.0.2.2:8065";
+    const url = `ws://${host}/api/v4/websocket`;
+    const origin = `https://${host}`;
```
josephbaylon, line 254: @migbot Quick question: should the protocol be `wss` instead of `https` ?

josephbaylon:
With the latest commit, API client requests for get/put/post/patch/delete/upload now crash as well.

Generic client requests for get/put/post/patch/delete and fast-image API client request work fine.

josephbaylon:
@migbot I ran the whole suite for regression testing and everything passed. I'll rerun the suite again after you push your self-signed certs fix and approve the PR if everything passes.

josephbaylon (approved):
LGTM. Regression testing passed. As discussed in chats, I'll test the certificate auth feature after merge to master, once I start creating automated e2e around it.

josephbaylon (approved):
LGTM. Regression testing passed. As discussed in chats, I'll test the certificate auth feature after merge to master, once I start creating automated e2e around it.

@types/APIClients.ts
```diff
@@ -1,94 +1,126 @@
...
-};
-
-interface GenericClientInterface {
-    get(url: string, options?: RequestOptions): Promise<ClientResponse>;
-    put(url: string, options?: RequestOptions): Promise<ClientResponse>;
-    post(url: string, options?: RequestOptions): Promise<ClientResponse>;
-    patch(url: string, options?: RequestOptions): Promise<ClientResponse>;
-    delete(url: string, options?: RequestOptions): Promise<ClientResponse>;
-}
+declare global {
```
shazm, line 6: I actually just had a thought around this - how about we declare `EmitterSubscription` and `NativeModule` as a global type in a `@types/Global.ts`, and that way we can use both of these without needing to import into our actual declarations.

Something like:

```
// @types/Global.ts

import type { NativeModule as NM, EventEmitter as EE } from "react-native";

declare global {
    interface NativeModule extends NM {}
    interface EventEmitter extends EE {}
}
```

This way we can use EventEmitter and NativeModule globally without needing to import the react types, thus not needing to do `declare global { ....` on our types.

@types/NativeAPIClient.ts
```diff
@@ -3,12 +3,22 @@
 import type { NativeModule } from "react-native";
 
 declare global {
-    type Constants = {
-        EXPONENTIAL_RETRY: "EXPONENTIAL_RETRY";
+    enum APIClientEvents {
```
shazm, line 6: Same as above - can potentially remove `declare global` here too.

shazm (changes requested):
Just a minor change which might clean up our type declarations; looks good otherwise! 👍

@types/APIClients.ts
migbot: @shazm OK pushed up some changes.

shazm (approved):

PR #45, opened 2021-03-09
Implement WebSocket client invalidation

migbot:
#### Summary
Adds `invalidateClient` to the WebSocket client which on iOS force disconnects the websocket, nils the delegate, removes the client from the manager, and deletes the identity stored in the Keychain if it was imported.


shazm (approved):
Seems to work fine for iOS, will do same for Android 👍

josephbaylon:
@migbot I forgot which PR this is waiting on (if any). Can this be merged as well?

josephbaylon (approved):
LGTM. I'm able to remove web socket client from example app

PR #50, opened 2021-03-18
[MM-31144] Android Uploads

shazm:
#### Summary
Allows android to upload both images and files via multi-part or as a direct stream. Both options read a file input stream directly, and write it to the request (either as a direct stream, or inside a multi-part).

Note: 
- UploadOptions now optionally allows `stream`, `skipBytes` and `method`; and upload screen updated
- Android Icons are fixed 🙌 

#### Ticket Link
Resolves MM-31144

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -7,6 +7,7 @@ import okio.*
 import java.util.*
 import com.mattermost.networkclient.enums.APIClientEvents
 import com.mattermost.networkclient.helpers.*
+import kotlin.collections.HashMap
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
@@ -195,7 +196,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     @Override
     override fun getConstants(): Map<String, Any> {
         val constants: MutableMap<String, Any> = HashMap<String, Any>()
-        constants["EXPONENTIAL_RETRY"] = "EXPONENTIAL_RETRY"
+        constants["RETRY_TYPES"] = hashMapOf("EXPONENTIAL_RETRY" to "exponential")
 
         // APIClient Events
         val events = HashMap<String, String>()
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -126,27 +126,45 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
 
     @ReactMethod
     fun upload(baseUrl: String, endpoint: String, file: String, taskId: String, options: ReadableMap?, promise: Promise) {
-        val skipBytes = if (options != null && options.hasKey("skipBytes")) options.getInt("skipBytes").toLong() else 0;
-        val stream = if (options != null && options.hasKey("stream")) options.getBoolean("stream") else false;
+        val body: RequestBody;
+        val uri = Uri.parse(file);
 
-        try {
-            // Create a Request Body
-            val body: RequestBody = if (stream) {
-                UploadFileRequestBody(reactApplicationContext, Uri.parse(file), skipBytes, ProgressListener(reactApplicationContext, taskId))
+        // check for multi-parts options
+        if(options != null && options.hasKey("multipart")){
+            val multipartOptions = options.getMap("multipart")!!
+
+            val multipartBody = MultipartBody.Builder();
+
+            if(multipartOptions.hasKey("fileKey") && multipartOptions.getString("fileKey") !== ""){
+                multipartBody.addFormDataPart(multipartOptions.getString("fileKey")!!, uri.lastPathSegment, UploadFileRequestBody(reactApplicationContext, uri, 0, ProgressListener(reactApplicationContext, taskId)))
             } else {
-                MultipartBody.Builder().addPart(UploadFileRequestBody(reactApplicationContext, Uri.parse(file), skipBytes, ProgressListener(reactApplicationContext, taskId))).build()
+                multipartBody.addPart(UploadFileRequestBody(reactApplicationContext, uri, 0, ProgressListener(reactApplicationContext, taskId)))
             }
 
+            if(multipartOptions.hasKey("data")){
+                val multipartData = multipartOptions.getMap("data")!!.toHashMap();
+                for( (k, v) in multipartData){
+                    multipartBody.addFormDataPart(k, v as String);
+                }
+            }
+
+            body = multipartBody.build()
+        } else {
+            val skipBytes = if (options != null && options.hasKey("skipBytes")) options.getInt("skipBytes").toLong() else 0;
+            body = UploadFileRequestBody(reactApplicationContext, Uri.parse(file), skipBytes, ProgressListener(reactApplicationContext, taskId))
+        }
+
+        try {
             // Create a Request
             var request = if(options?.hasKey("method") == true) {
                 when (options.getString("method")) {
-                    "PUT" -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").put(body)
-                    "PATCH" -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").patch(body)
+                    "PUT" -> sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).put(body)
+                    "PATCH" -> sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).patch(body)
                     // Default to "POST"
-                    else -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body)
+                    else -> sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).post(body)
                 }
             } else {
-                sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body)
+                sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).post(body)
             }
 
             // Parse options into the request / client
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -138,7 +138,16 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
             }
 
             // Create a Request
-            var request = sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body);
+            var request = if(options?.hasKey("method") == true) {
+                when (options.getString("method")) {
+                    "PUT" -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").put(body)
+                    "PATCH" -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").patch(body)
+                    // Default to "POST"
+                    else -> sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body)
+                }
+            } else {
+                sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body)
+            }
 
             // Parse options into the request / client
             if (options != null) request = request.parseOptions(options, sessionsClient[baseUrl]!!)
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -1,16 +1,18 @@
 package com.mattermost.networkclient
 
+import android.net.Uri
 import com.facebook.react.bridge.*
-import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okio.IOException
+import okhttp3.*
+import okio.*
 import java.util.*
+import com.mattermost.networkclient.enums.APIClientEvents
+import com.mattermost.networkclient.helpers.*
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
     var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>()
     var sessionsRequest = mutableMapOf<String, Request.Builder>()
+    var calls = mutableMapOf<String, Call>()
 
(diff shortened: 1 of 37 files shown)


android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -26,6 +28,9 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
             // Attach client options if they are passed in
             sessionsClient[baseUrl]!!.parseOptions(options, sessionsRequest[baseUrl]);
 
+            // Protocols
+            sessionsClient[baseUrl]!!.protocols(listOf(Protocol.HTTP_1_1))
```
enahum, line 32: Why 1.1? Will this work with http2?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -122,10 +127,56 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+        try {
+            // Create a Request Body
+            val body: RequestBody = if (stream) {
+                UploadFileRequestBody(reactApplicationContext, Uri.parse(file), skipBytes, ProgressListener(reactApplicationContext, taskId))
+            } else {
+                MultipartBody.Builder().addPart(UploadFileRequestBody(reactApplicationContext, Uri.parse(file), skipBytes, ProgressListener(reactApplicationContext, taskId))).build()
+            }
+
+            // Create a Request
+            var request = sessionsRequest[baseUrl]!!.url("$baseUrl/$endpoint").post(body);
```
enahum, line 144: Can't the method be put and or patch? Is it true that this must be a post method at all times?

android/src/main/java/com/mattermost/networkclient/helpers/UploadProgressListener.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    override fun emitProgressEvent(progress: Double) {
+        val params = Arguments.createMap()
+        params.putDouble("fractionCompleted", progress)
+        params.putString("taskId", "$taskId")
+        reactContext.getJSModule(RCTDeviceEventEmitter::class.java).emit(APIClientEvents.UPLOAD_PROGRESS.event, params)
+    }
+
+    override fun update(bytesRead: Double, contentLength: Double) {
+        val progress = bytesRead / contentLength;
```
enahum, line 24: As we are passing the % value to JS, should we round the value? Wouldn't be better to send both values to JS and let the consumer use those values as they see fit? @migbot thoughts?

android/src/main/java/com/mattermost/networkclient/helpers/UploadRequestBody.kt
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException
+
+
+class UploadFileRequestBody(private val reactContext: ReactContext, private val uri: Uri, private val skipBytes: Long, private val listener: ProgressListener) : RequestBody() {
+
+    private val stream = reactContext.contentResolver.openInputStream(uri)!!
+    private val total: Double = stream.available().toDouble();
+
+    override fun contentLength(): Long {
+        return total.toLong();
```
enahum, line 20: Is this value always available?

android/src/main/java/com/mattermost/networkclient/helpers/UploadRequestBody.kt
```diff
@@ -0,0 +1,50 @@
...
+                totalRead = skipBytes;
+            }
+
+            while (source.read(sink.buffer, 65536L).also { read = it } != -1L) {
+                sink.flush()
+                totalRead += read
+                listener.update(totalRead.toDouble(), total)
+            }
+        } catch (e: IOException) {
+            source.closeQuietly()
```
enahum, line 45: So when this fails how do we know that it failed?

Also finally executes even after the catch, so the same code is running twice here

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/helpers/UploadRequestBody.kt
shazm: Whoops - this is supposed to throw further upstream where it can be caught / handled.

android/src/main/java/com/mattermost/networkclient/helpers/UploadRequestBody.kt
shazm: Yup; I grab the stream's length and put it in the `total` val; though that's assuming there is a stream - otherwise it returns `content-length: 0`

android/src/main/java/com/mattermost/networkclient/helpers/UploadProgressListener.kt
shazm: I think the JS side is expecting a fraction between 1 and 0; and it can then workout what it wants to do with it

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Yup it can be! Should we allow for any of the three and just leave `post` as default?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Whoops! Was testing against various servers on my local - removed.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
enahum: Yup, we should allow the method to be defined and fallback to post if undefined

android/src/main/java/com/mattermost/networkclient/helpers/UploadProgressListener.kt
enahum: What if I want to display the progress not in percentage? Every other lib sends both values in kb and you can calculate the % as you wish. We can send a double from 0.00 to 1.00 but then make sure the fraction only have two decimals

android/src/main/java/com/mattermost/networkclient/helpers/UploadRequestBody.kt
enahum: So make sure you don't divide by 0

android/src/main/java/com/mattermost/networkclient/helpers/UploadProgressListener.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun emitProgressEvent(progress: Double) {
+        val params = Arguments.createMap()
+        params.putDouble("fractionCompleted", progress)
+        params.putString("taskId", "$taskId")
+        reactContext.getJSModule(RCTDeviceEventEmitter::class.java).emit(APIClientEvents.UPLOAD_PROGRESS.event, params)
+    }
+
+    override fun update(bytesRead: Double, contentLength: Double) {
+        // Only emit if we can show progress against content length
+        if(contentLength > 0) emitProgressEvent(df.format(bytesRead / contentLength).toDouble())
```
shazm, line 28: @enahum / @migbot - We would still want to show some progress on the upload bar no? I'm wondering if there's ever a scenario where it's unable to compute the length of the stream but a stream still exists that it starts uploading...

enahum: If no content lenght I would return the bytesRead..

josephbaylon:
@shazm As discussed in the chats,
1. I pushed some e2e fixes for iOS. I also disabled upload e2e for Android because of automated file attachment limitation.
2. Android can only be tested manually for now. Stream upload throws error saying stream is closed. Will only test multi-part for Android
3. Separate PR for fast image fixes

@migbot For parity, I'm guessing iOS will also implement multi-part and skip-bytes for stream.

josephbaylon:
@shazm I pushed some new changes so that existing file server can accommodate multipart request. I tried it locally and it's able to receive request from postman. However, from the example app, it times out. Can you please run the android example on your end and the try the file server?

1. Run under example `npm run file-upload-server`
2. Use path: `/api/files/multipart`

Here are the snapshots from postman showing headers and body and successful upload.
|Headers|Body|
|:---|:---|
|Screen Shot 2021-03-19 at 11 35 17 AM|Screen Shot 2021-03-19 at 11 34 57 AM|

Here's the snapshot of the android app timeout
Screen Shot 2021-03-19 at 11 54 29 AM

android/src/main/java/com/mattermost/networkclient/enums/APIClientEvents.kt
```diff
@@ -0,0 +1,6 @@
+package com.mattermost.networkclient.enums
+
+enum class APIClientEvents(val event: String) {
+    UPLOAD_PROGRESS("APIClient-UploadProgress"),
+    CLIENT_CERTIFICATE_MISSING("APIClient-Client-Certificate-Missing"),
```
migbot, line 5: FYI I've removed this [here](https://github.com/mattermost/react-native-network-client/pull/52/files#diff-ceeeb3e5b18acce6df7a9516ad89c5babbf12e2204a2cc41569582fc0678afbeR35) in favor of a general error event that will be sent when the client throws a client cert missing error.

example/src/screens/CreateAPIClientScreen.tsx
```diff
@@ -80,7 +80,7 @@ export default function CreateAPIClientScreen({
             requestAdapterConfiguration,
         };
 
-        if (clientP12Configuration.path) {
+        if (clientP12Configuration.path && clientP12Configuration.path !== "") {
```
migbot, line 83: Unnecessary since empty string is falsey.

android/src/main/java/com/mattermost/networkclient/helpers/UploadProgressListener.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class ProgressListener(private val reactContext: ReactContext, private val taskId: String) : ProgressListenerInterface {
+    // 2-Decimal places, rounded up
+    private val df = DecimalFormat("#.00", ).apply{ roundingMode = RoundingMode.UP }
+
+    override fun emitProgressEvent(progress: Double, bytesRead: Double) {
+        val params = Arguments.createMap()
+        params.putDouble("fractionCompleted", progress)
+        params.putString("taskId", "$taskId")
+        params.putDouble("bytesRead", bytesRead)
```
migbot, line 23: I thought we wanted to send `bytesRead` and `contentLength` instead of `fractionCompleted`?

shazm: Oh I'm just going off: https://github.com/mattermost/react-native-network-client/pull/50#discussion_r597286263

I figure both and the JS side should have the size already as it plucked the file; that way they can use either the fraction or the bytes read and workout the progress themselves; plus negates the chance where we are unable to read the content-length of the stream.

migbot: Ah right right!

josephbaylon (changes requested):
Please see issues found here https://github.com/mattermost/react-native-network-client/pull/50#issuecomment-803044201

shazm:
Hey @josephbaylon - I think the issues have been resolved, but we've introduced a few new things:

- Default upload type is "stream", with multi-part being optional
- If multi-part is selected, then you can optionally provide a "key" for the file (@enahum pointed out how MM API calls for `files` on the file upload endpoint, but `image` on the user profile picture upload endpoint)
- You can then also pass in optional `key: string = value: string` parts to the multi-part as well
- `skipBytes` should only be for stream uploads, and optional
- Uploads should be able to call post / put / patch as required

^^ I think that changes / touches pretty much every test; and I think @migbot is doing the equivalent on iOS

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -122,10 +124,83 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+        val uri = Uri.parse(file);
+
+        // check for multi-parts options
+        if(options != null && options.hasKey("multipart")){
+            val multipartOptions = options.getMap("multipart")!!
+
+            val multipartBody = MultipartBody.Builder();
+
+            if(multipartOptions.hasKey("fileKey") && multipartOptions.getString("fileKey") !== ""){
+                multipartBody.addFormDataPart(multipartOptions.getString("fileKey")!!, uri.lastPathSegment, UploadFileRequestBody(reactApplicationContext, uri, 0, ProgressListener(reactApplicationContext, taskId)))
```
migbot, line 139: Non-blocker but this might be easier read without nested calls.

migbot (approved):

josephbaylon:
@shazm I'm now getting this error when building android.

Screen Shot 2021-03-23 at 8 55 38 AM

enahum:
@shazm can I ask you that before submitting a PR or a new commit, test it locally and ensure that it works to avoid going back and forth all the time, this will help speed up things.

Tests should be manual and by running the E2E tests we already have.

🍀

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -122,10 +125,83 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+            } else {
+                multipartBody.addPart(UploadFileRequestBody(reactApplicationContext, uri, 0, ProgressListener(reactApplicationContext, taskId)))
```
enahum, line 142: I thought we decided to use "files" as the default key

enahum (changes requested):
Won't be reviewing the example app code. As far as the lib goes all looks good but the default key for multi-part uploads

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: Ah whoops - will change.

josephbaylon:
@shazm File upload for stream works but for multipart I'm still getting a timeout error. Can you please test locally and use the preset `File Upload Server API` client and 

1. Run under example npm run file-upload-server
2. Use path: /api/files/multipart

shazm:
hey @josephbaylon - I did a lot of digging and the `multipartUploadHandler` seemed to be crashing for some reason but it wasn't logging out; i wrapped it in a try/catch and changed the `for-loop` to an `Object.entries().map()` which seems to work now for android. Will double check if it breaks anything in iOS a bit later today.

josephbaylon (approved):
@shazm Thanks for fixing the multipart!! Both stream and multipart upload now work on Android. Approving this PR!

enahum (approved):

PR #47, opened 2021-03-11
[MM-22753, MM-30123] iOS - Adds a linear retry policy and returns `retriesExhausted` in the response

migbot:
#### Summary
Adds a second retry policy of `linear` and returns `retriesExhausted` in the response payload when applicable.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22753
https://mattermost.atlassian.net/browse/MM-30123

PR #52, opened 2021-03-19
MM-31773 Return appropriate errors/warnings from iOS to JS

migbot:
#### Summary
* Emit warning events for errors that do not necessitate rejection of promises (ie, invalid password of p12 during creation of API client)
* Use existing iOS errors for malformed URL and values not found
* Update native reject calls to include correct code, message, domain, userInfo, and nativeStackAndroid/nativeStackIOS
* Reject on retries exhausted
* For error codes I chose `-1xx` range for Keychain errors, `-2xx` for APIClient errors, and `-3xx` for general NetworkClient errors.

Example general rejection:
example_rejection

Example rejection due to retries exhausted:
retries_exhausted_rejection

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31773


src/APIClient/index.tsx
```diff
@@ -57,6 +58,16 @@ class APIClient implements APIClientInterface {
                 }
             }
         );
+        this.clientWarningSubscription = Emitter.addListener(
+            EVENTS.WARNING,
+            (event: APIClientWarningEvent) => {
+                if (event.serverUrl === this.baseUrl) {
+                    Alert.alert("Warning", event.warning, [{ text: "OK" }], {
```
enahum, line 65: Why display an alert here? I guess this should be fwd to the applicstion making use of the lib (aka the example app) and let it handle the UX as they see fit

src/WebSocketClient/index.tsx
```diff
@@ -18,19 +18,32 @@ const SOCKETS: { [key: string]: WebSocketClient } = {};
...
                     this.readyState = event.message as WebSocketReadyState;
                 }
             }
         );
+        this.clientWarningSubscription = Emitter.addListener(
+            // TODO: remove subscription on invalidate
+            EVENTS.WARNING,
+            (event: WebSocketClientWarningEvent) => {
+                if (event.url === this.url) {
+                    Alert.alert("Warning", event.warning, [{ text: "OK" }], {
```
enahum, line 41: Same here

enahum (changes requested):
Apart from my comments, why are the events warnings and not errors? And why do we send the localize message only? Perhaps a code would be good instead of trying to match localized strings if the app wants to handle different UX based on the code.

I can see how MM would use the cert missing "warning" to provide some UI to import the cert as a response

src/APIClient/index.tsx
migbot: Oh duh, sorry that was the intention! Will fix.

src/APIClient/index.tsx
```diff
@@ -37,39 +37,38 @@ const generateUploadTaskId = () =>
...
+            EVENTS.CLIENT_ERROR,
+            (event: APIClientErrorEvent) => {
                 if (event.serverUrl === this.baseUrl) {
-                    Alert.alert("Warning", event.warning, [{ text: "OK" }], {
-                        cancelable: false,
-                    });
+                    if (this.clientErrorEventHandler) {
+                        this.clientErrorEventHandler(event);
+                    } else {
+                        Alert.alert(
```
migbot, line 54: @enahum I added this `else` to avoid the risk of missing errors if a consumer does not register a `clientErrorEventHandler`.

shazm:
> Apart from my comments, why are the events warnings and not errors? And why do we send the localize message only? Perhaps a code would be good instead of trying to match localized strings if the app wants to handle different UX based on the code.
> 
> I can see how MM would use the cert missing "warning" to provide some UI to import the cert as a response

Oh so the thinking was - if there is an "error", the called promise should be rejected and handled accordingly; for a lot of the places we're emitting events we still want the promise to resolve (even if partially) and just return a message / warning to the user. 

The main example is trying to make an API client with a certificate; if there is an issue with the certificate i.e. wrong password, the client should still be created but a warning to the user given, and then the option to add the certificate later to that client.

shazm (approved):
Looks good to me - just would like to clean up our naming around host/baseUrl/domain.

@types/APIClients.ts
```diff
@@ -31,6 +31,17 @@ type ClientResponse = {
     retriesExhausted?: boolean;
 };
 
+type ClientResponseError = {
+    code: number;
+    message: string;
+    domain: string;
```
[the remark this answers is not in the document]
migbot: Ah `domain` here is an [iOS NSURLErrorDomain](https://developer.apple.com/documentation/foundation/nsurlerrordomain) :D

@types/APIClients.ts
```diff
@@ -31,6 +31,17 @@ type ClientResponse = {
...
+    userInfo?: Record<string, unknown>;
+    nativeStackAndroid?: Array<unknown>;
+    nativeStackIOS?: Array<unknown>;
```
enahum, line 40: why the difference?

@types/WebSocketClient.ts
```diff
@@ -19,9 +19,19 @@ type WebSocketEvent = {
...
+    event: WebSocketClientErrorEvent
+) => void;
+
 interface WebSocketClientInterface {
     url: string;
     readyState: WebSocketReadyState;
+
+    webSocketEventSubscription: EmitterSubscription;
+    clientErrorSubscription: EmitterSubscription;
+    clientErrorEventHandler?: WebSocketClientErrorEventHandler;
```
enahum, line 32: bad copy/paste?

src/APIClient/index.tsx
enahum: So in case this is unhandled because I don't want need to handle it you are forcing me to get an Alert? I rather remove the else statement

src/WebSocketClient/index.tsx
```diff
@@ -18,19 +18,40 @@ const SOCKETS: { [key: string]: WebSocketClient } = {};
...
             }
         );
+        this.clientErrorSubscription = Emitter.addListener(
+            EVENTS.CLIENT_ERROR,
+            (event: WebSocketClientErrorEvent) => {
+                if (event.url === this.url) {
+                    if (this.clientErrorEventHandler) {
+                        this.clientErrorEventHandler(event);
+                    } else {
+                        Alert.alert(
```
enahum, line 43: same here

enahum (changes requested):
Please add the error codes to the README.md file :)

src/APIClient/index.tsx
migbot: Ah right, OK will remove.

@types/APIClients.ts
migbot: This is what RN will return in the rejection. See [here for Android](https://github.com/facebook/react-native/blob/6e6443afd04a847ef23fb6254a84e48c70b45896/ReactAndroid/src/main/java/com/facebook/react/bridge/PromiseImpl.java#L29-L32) and [here of IOS](https://github.com/facebook/react-native/blob/0f4f91766339f2cf4189446d7d575493d33b4009/React/Base/RCTUtils.m#L478-L511). Do you want me to parse to a consistent structure? I was thinking that it's best to leave as is so that a consumer can handle rejections here the same way they would in other RN code.

@types/APIClients.ts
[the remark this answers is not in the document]
migbot: See my [comment](https://github.com/mattermost/react-native-network-client/pull/52#discussion_r598901062) below.

@types/WebSocketClient.ts
migbot: Looks right to me, what do you mean?

enahum: Got confused with clientError and WSClientError

@types/APIClients.ts
```diff
@@ -31,6 +31,17 @@ type ClientResponse = {
     retriesExhausted?: boolean;
 };
 
+type ClientResponseError = {
+    code: number;
+    message: string;
+    domain: string;
+    userInfo?: Record<string, unknown>;
```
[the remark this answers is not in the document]
migbot: I was hoping to find some type information in the RN source code but only found [userInfo in relation to notifications](https://github.com/facebook/react-native/blob/f1f9cef9d331162bd7a5c98ac5c5fdded9a24b78/Libraries/PushNotificationIOS/NativePushNotificationManagerIOS.js#L26). I'm not too familiar yet with generics. How would `T` be used here?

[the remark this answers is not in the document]
enahum: You can read more about it in [this blog post](https://rossbulat.medium.com/typescript-generics-explained-15c6493b510f) I was thinking that each consumer can define the type they expect to get back to make it less error prone

README.md
```diff
@@ -11,15 +11,69 @@ npm install react-native-network-client
...
+  "https://community.mattermost.com"
+);
+const { client: wsClient, created } = await getOrCreateWebSocketClient(
+  "wss://community.mattermost.com"
+);
 ```
 
+## Errors
+
+There are two cases where errors will be returned by the native code. The first is via rejected promises and the second is via events.
```
enahum, line 32: Isn't the second one really using an error callback triggered by the JS side of the lib? I tought only the Lib was listening for the events, maybe I'm wrong?

README.md
```diff
@@ -11,15 +11,69 @@ npm install react-native-network-client
...
+
+In the case of rejected promises, the error structure is unchanged from what React Native returns:
+
+```
+{
+    code: number;
+    message: string;
+    userInfo: Object;
```
enahum, line 40: Ueah maybe just call this extra. And about domain, is it really needed?

README.md
```diff
@@ -11,15 +11,69 @@ npm install react-native-network-client
...
+    errorCode: number;
+    errorDescription: string;
+}
+```
+
+and you'll need to subscribe to these events by passing your error handler as a callback to `onClientError(callback: WebSocketClientErrorEventHandler): void;`.
+
+For both API client and WebSocket client, the following error codes apply:
+
+- Keychain specific errors will be in the range -100 to -199
```
enahum, line 73: Maybe explain that this are related to client certs

enahum (changes requested):
Also consider adding a table with a list of error codes and descriptions

josephbaylon:
@migbot It seems you're using a code formatter. Can you share that so I can use that locally as well?

josephbaylon:
@migbot I know this is iOS but getting the following error when building android using `npm run android`

Screen Shot 2021-03-22 at 12 36 36 PM

README.md
migbot: Not sure I understand the question, but for most all events we are sending back we expect the consumer to register callbacks for via the proper client functions. The only exception is for the WebSocket ready state where we register the listener in the contructor.

And actually, I realized I could refactor the subscription handling bit better. See latest commit.

josephbaylon:
The issue https://github.com/mattermost/react-native-network-client/pull/52#issuecomment-804341145 looks like it was introduced by https://github.com/mattermost/react-native-network-client/pull/47 and android testing was missed there.

@migbot @shazm I'll let you guys decide who will fix the issue or if it's a separate PR. I'll go ahead and approve the PR based on running the current test suite. I already created jira tickets to create e2e for errors/warnings once this is merged.

josephbaylon (approved):
LGTM. iOS e2e passed locally

README.md
migbot: I'd rather just return the structure untouched so it's consistent with any RN rejection.

enahum (approved):
I guess the table in the readme needs to be updated with Android as well, right now it is entirely ios focused

PR #53, opened 2021-03-19
[MM-30124, MM-34747] Retries for Android and iOS

shazm:
#### Summary

Introduces the following changes to Library:

- Setting retryable HTTP Status Codes 
- Setting retryable HTTP Methods
- Example App updated via RequestConfiguration to handle the above

Android specific changes:
- Android / Kotlin Clients / Requests updated to use a singleton pattern; thus allowing configuration to be held / accessed outside of the main library class
- Base URL's are parsed using `.toHttpUrl().toString()` to provide consistency 
- Timeouts updated to Milliseconds to reflect iOS usage
- Default config params for retries are used if none are available on the client or request
- All responses are resolved successfully via `promise.resolve()`, only non-responses are thrown as exceptions

iOS:
- WIP on adding HTTP Status Codes / Methods (for @migbot )

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30124
https://mattermost.atlassian.net/browse/MM-34747

### DIFF
--- a/@types/APIClients.ts
+++ b/@types/APIClients.ts
@@ -103,6 +103,8 @@ type RetryPolicyConfiguration = {
     retryInterval?: number;
     exponentialBackoffBase?: number;
     exponentialBackoffScale?: number;
+    statusCodes?: number[];
+    retryMethods?: string[];
 };
 
 type RequestAdapterConfiguration = {
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -63,6 +63,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     fun invalidateClientFor(baseUrl: String, promise: Promise) {
         try {
             sessionsRequest.remove(baseUrl.trimSlashes());
+            sessionsClient.remove(baseUrl.trimSlashes());
             promise.resolve(sessionsRequest.keys);
         } catch (err: Throwable) {
             promise.reject(err)
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -13,9 +13,8 @@ import kotlin.collections.HashMap
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
     private val sessionsClient = SessionsObject.client
-    private val sessionsRequest = SessionsObject.request
     private val sessionsCall = SessionsObject.call
-    private val sessionsConfig = SessionsObject.config
+    private val sessionsConfig = SessionsObject.requestConfig
 
     override fun getName(): String {
         return "APIClient"
@@ -24,16 +23,15 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     @ReactMethod
     fun createClientFor(baseUrl: String, options: ReadableMap, promise: Promise) {
         // Don't trust user input...
-        val url = baseUrl.toHttpUrl().toString();
+        val url = baseUrl.toUrlString()
 
         try {
             // Create the client and request builder
             sessionsClient[url] = OkHttpClient().newBuilder();
-            sessionsRequest[url] = Request.Builder().url(url);
             sessionsConfig[url] = hashMapOf("baseUrl" to url);
 
             // Attach client options if they are passed in
-            sessionsClient[url]!!.parseOptions(options, sessionsRequest[url], url);
+            sessionsClient[url]!!.applyClientOptions(options, url);
 
             promise.resolve(null)
         } catch (err: Throwable) {
@@ -46,8 +44,17 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         // Don't trust user input...
         val url = baseUrl.toHttpUrl().toString();
 
+        val headers = SessionsObject.requestConfig[url]!!["clientHeaders"] as ReadableMap?
+        val map = Arguments.createMap();
+
+        if(headers != null){
+            for((k, v) in headers.toHashMap()){
+                map.putString(k, v as String)
+            }
+        }
+
         try {
-            promise.resolve(sessionsRequest[url]?.build()?.headers?.readableMap())
+            promise.resolve(map)
         } catch (error: Error) {
             promise.reject(error)
         }
@@ -59,7 +66,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            sessionsRequest[url]?.addHeadersAsReadableMap(headers)
+            SessionsObject.requestConfig[url]!!["clientHeaders"] = headers
             promise.resolve(null);
         } catch (error: Error) {
             promise.reject(error)
@@ -72,10 +79,9 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            sessionsRequest.remove(url);
             sessionsClient.remove(url);
             sessionsConfig.remove(url);
-            promise.resolve(sessionsRequest.keys);
+            promise.resolve(sessionsClient.keys);
         } catch (err: Throwable) {
             promise.reject(err)
         }
@@ -87,10 +93,17 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).parseOptions(options, sessionsClient[url]!!, url).build();
-            sessionsClient[url]!!.build().newCall(request).execute().use { response ->
-                promise.resolve(response.returnAsWriteableMap(url))
-            }
+            val request = Request.Builder()
+                    .url(formUrlString(url, endpoint))
+                    .applyClientOptions(url)
+                    .applyRequestOptions(options, url)
+                    .build();
+            sessionsClient[url]!!
+                    .build()
+                    .newCall(request)
+                    .execute().use { response ->
+                        promise.resolve(response.returnAsWriteableMap(url))
+                    }
         } catch (e: IOException) {
             promise.reject(e)
         }
@@ -102,7 +115,14 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).post(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
+//            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).post(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
+
+            val request = Request.Builder()
+                    .url(formUrlString(url, endpoint))
+                    .applyClientOptions(url)
+                    .applyRequestOptions(options, url)
+                    .post(options.getMap("body")!!.bodyToRequestBody())
+                    .build();
             sessionsClient[url]!!.build().newCall(request).execute().use { response ->
                 promise.resolve(response.returnAsWriteableMap(url))
             }
@@ -117,7 +137,12 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).put(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
+            val request = Request.Builder()
+                    .url(formUrlString(url, endpoint))
+                    .applyClientOptions(url)
+                    .applyRequestOptions(options, url)
+                    .put(options.getMap("body")!!.bodyToRequestBody())
+                    .build();
             sessionsClient[url]!!.build().newCall(request).execute().use { response ->
                 promise.resolve(response.returnAsWriteableMap(url))
             }
@@ -132,7 +157,13 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).patch(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
+            val request = Request.Builder()
+                    .url(formUrlString(url, endpoint))
+                    .applyClientOptions(url)
+                    .applyRequestOptions(options, url)
+                    .patch(options.getMap("body")!!.bodyToRequestBody())
+                    .build();
+
             sessionsClient[url]!!.build().newCall(request).execute().use { response ->
                 promise.resolve(response.returnAsWriteableMap(url))
             }
@@ -147,7 +178,12 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
         val url = baseUrl.toHttpUrl().toString();
 
         try {
-            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).delete(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
+            val request = Request.Builder()
+                    .url(formUrlString(url, endpoint))
+                    .applyClientOptions(url)
+                    .applyRequestOptions(options, url)
+                    .delete(options.getMap("body")!!.bodyToRequestBody())
+                    .build();
             sessionsClient[url]!!.build().newCall(request).execute().use { response ->
                 promise.resolve(response.returnAsWriteableMap(url))
             }
@@ -194,17 +230,17 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
             // Create a Request
             var request = if (options?.hasKey("method") == true) {
                 when (options.getString("method")) {
-                    "PUT" -> sessionsRequest[url]!!.url(formUrlString(url, endpoint)).put(body)
-                    "PATCH" -> sessionsRequest[url]!!.url(formUrlString(url, endpoint)).patch(body)
+                    "PUT" -> Request.Builder().url(formUrlString(url, endpoint)).put(body)
+                    "PATCH" -> Request.Builder().url(formUrlString(url, endpoint)).patch(body)
                     // Default to "POST"
-                    else -> sessionsRequest[url]!!.url(formUrlString(url, endpoint)).post(body)
+                    else -> Request.Builder().url(formUrlString(url, endpoint)).post(body)
                 }
             } else {
-                sessionsRequest[url]!!.url(formUrlString(url, endpoint)).post(body)
+                Request.Builder().url(formUrlString(url, endpoint)).post(body)
             }
 
             // Parse options into the request / client
-            if (options != null) request = request.parseOptions(options, sessionsClient[url]!!, url)
+            if (options != null) request = request.applyClientOptions(url).applyRequestOptions(options, url)
 
             // Create a cancellable call
             sessionsCall[taskId] = sessionsClient[url]!!.build().newCall(request.build())
--- a/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
@@ -10,9 +10,10 @@ import kotlin.collections.HashMap
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
 
-    var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>()
-    var sessionsRequest = mutableMapOf<String, Request.Builder>()
-    var calls = mutableMapOf<String, Call>()
+    private val sessionsClient = SessionsObject.client
+    private val sessionsRequest = SessionsObject.request
+    private val sessionsCall = SessionsObject.call
+    private val sessionsConfig = SessionsObject.config
 
     override fun getName(): String {
         return "APIClient"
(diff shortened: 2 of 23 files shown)


shazm:
OkHttp3 doesn't support per-request interceptors; so being able to attach a `RetryPolicyConfiguration` to a request isn't going to work. It works perfectly with the settings used on the client on initialisation.

I've also added a small `retry_server.js` that returns `418` for the first 5 calls and then `200` for the 6th one while logging the time diff to check the linear and exponential wait times.

josephbaylon (changes requested):
@shazm When I tested on Android, I get a connection error (see images below)

Steps,
1. Run retry server `node retry_server.js`
2. Make sure it works on postman
3. Generic client --> http://localhost:8000
4. Make request

Expected: Response overlay with 418 code
Actual: Crash error

Postman:
Screen Shot 2021-03-29 at 3 39 12 PM

Client setup:
Screen Shot 2021-03-29 at 3 35 32 PM

Client request crash:
Screen Shot 2021-03-29 at 3 37 04 PM

josephbaylon:
Also note that when I create E2E for client retries, I'll incorporate the retry mechanism to the new in-house mockserver (https://github.com/mattermost/react-native-network-client/blob/6fc40d4e1956e4c0a4425c62bb905cef378fd2f1/example/servers/mockserver.js) so that retry is not a separate server. Client will just need to pass retry param, e.g. `?retry=5`

enahum:
> Also note that when I create E2E for client retries, I'll incorporate the retry mechanism to the new in-house mockserver (https://github.com/mattermost/react-native-network-client/blob/6fc40d4e1956e4c0a4425c62bb905cef378fd2f1/example/servers/mockserver.js) so that retry is not a separate server. Client will just need to pass retry param, e.g. `?retry=5`

@josephbaylon how do yo expect the client to pass the retry flag, is it expected to bump on each retry?

josephbaylon:
@shazm Regarding the failing screen https://github.com/mattermost/react-native-network-client/pull/53#pullrequestreview-623707525 it's totally a user error

I used `10.0.2.2` and it now returns the 418 code

josephbaylon:
@enahum It's just rough idea now but based on the retry server that @shazm created, he has retry attempts hardcoded to 5, so maybe that can be passed in as parameter and also a delay interval for server that can be controlled through the test. I haven't really brainstormed on it much (of course it will change later as I setup the test).

As for passing the parameters it will be part of the url, e.g. http://localhost:8000?retry=5&delay=2000

josephbaylon (approved):
LGTM. Approving this for now. I will conduct further testing when I get around creating the e2e for retries.

https://mattermost.atlassian.net/browse/MM-34433

enahum:
@josephbaylon 
I don't see how it will differentiate (the server) if it is retry # 1 or 5 if it is passed in the url. Retries should happen automatically and without changing the uri, so hardcoded for testing seems like a good alternative for me

enahum:
@shazm can you update the description to point the link to the PR this is based from instead of the branch

josephbaylon:
@enahum I'm thinking more like creating a response delay from the server (delay = response delay, retryLimit = server retry limit), something like

const delay = getResponseDelay(req);
Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, delay);
if (attempts > serverRetryLimit) ...

so that the client will actually attempt to retry. 
if server delay > client retry limit * interval then client should retry.


But as I'll have a more concrete example once I work the e2e

shazm:
> @josephbaylon
> I don't see how it will differentiate (the server) if it is retry # 1 or 5 if it is passed in the url. Retries should happen automatically and without changing the uri, so hardcoded for testing seems like a good alternative for me

Oh I think this not which retry attempt it is, but how many times to return a 4XX before returning a 2XX.

So if I understand @josephbaylon correctly, in the `APIRetryClient` for testing I would call something like `localhost:8000?retryLimit=10&delay=500` - so the server then knows it needs to return 4XX for the next 9 calls (assuming current call will be 4XX too) before returning 2XX.

Currently the `retry_server.js` is just hard-coded to 5; so the 6th call will always be 2XX.

enahum:
> > @josephbaylon
> 
> > I don't see how it will differentiate (the server) if it is retry # 1 or 5 if it is passed in the url. Retries should happen automatically and without changing the uri, so hardcoded for testing seems like a good alternative for me
> 
> 
> 
> Oh I think this not which retry attempt it is, but how many times to return a 4XX before returning a 2XX.
> 
> 
> 
> So if I understand @josephbaylon correctly, in the `APIRetryClient` for testing I would call something like `localhost:8000?retryLimit=10&delay=500` - so the server then knows it needs to return 4XX for the next 9 calls (assuming current call will be 4XX too) before returning 2XX. I'm assuming the server will be testing for the delay (making sure calls come in every 500ms; might want to have a 5 or 10% buffer on either side for network/latency/processing?). Also assuming that subsequent query params will be ignored until 2XX is returned at which point you can reset.
> 
> 
> 
> Currently the `retry_server.js` is just hard-coded to 5; so the 6th call will always be 2XX.

Exactly, so if every retried request has the same retryLimit value, how does the server keep track that it has returned a 4XX call retryLimit times +/- one?

shazm:
> Exactly, so if for every retried request has the same retryLimit value, how does the server keep track that it has returned a 4XX call retryLimit times +/- one?

Oh it's a single thread / dumb server - at least the way I made it and I think what @josephbaylon was alluding to; the first time it sees `retryLimit` it just sets that value and then increments the internal counter for every request it gets until `req > retryLimit` at which point it just returns 2XX. The test should be complete by then and the server reset / restarted for the next one (if there will be another).

Alternatively, we can just have a few end points and hard coded values? So something like:

`localhost:8000/retry-5` :: Keeps an internal counter to 5, after which it returns 2XX
`localhost:8000/retry-10` :: Keeps an internal counter to 10, after which it returns 2XX
`localhost:8000/reset` :: Resets all counters and so retries revert to returning 4XX

This way we can have a long-living server process?

enahum:
You guys decide what's best

josephbaylon:
@enahum I think that can be achieved by supplying a unique id to url and pushing it to map variable in the server that keeps track of the retryLimit value,

e.g,
```
if (!serverRetryMap.has(req.uniqueTestID)) {
    serverRetryMap.push(req.uniqueTestID, {serverDelay: req.serverDelay, serverRetryLimit: req.serverRetryLimit, clientAttempts: 1});
} else {
    const serverRetry = serverRetryMap.get(req.uniqueTestID);
    ....
}
```

josephbaylon:
@shazm I'll modify the `retry_server.js` sometime tomorrow when I'm free again and push it to this PR, so that it has the capability of what we're trying to achieve

android/src/main/java/com/mattermost/networkclient/interceptors/RetryInterceptor.kt
```diff
@@ -3,39 +3,59 @@ package com.mattermost.networkclient.interceptors
...
+        private var exponentialBackOffBase: Double?,
+        private var exponentialBackOffScale: Double?
 ) : Interceptor {
 
-    private val type = _type ?: "EXPONENTIAL";
-    private val retryLimit = _retryLimit ?: 10;
-    private val exponentialBackOffBase = _exponentialBackOffBase ?: 2.0
-    private val exponentialBackOffScale = _exponentialBackOffScale ?: 0.5
+    init {
+        if (type == null) type = "EXPONENTIAL"
```
migbot, line 19: Why default instead of just foregoing the interceptor if type is null?

josephbaylon:
@shazm I updated `example/retry_server.js` so that it can receive query parameters and delay response based on params

e.g.
```
http://localhost:8000/?clientID=123&serverDelay=10000&serverRetryLimit=5
```

My attempt to test on generic client or api client results in app crashing, but the same on postman works. Maybe you could play around with it and update if necessary.

shazm:
Ah whoops - I didn't mean to push up my last commit; that's definitely a WIP for the singleton pattern change, fixing it up now.

shazm:
@enahum / @josephbaylon / @migbot - Just corrected a lot of small things on this and per-request config is now available.

Note: I've moved out the client(s), request(s), config(s) and call(s) into their own singleton objects, and I check to see if there's a retry policy first in the request, then in the client, and lastly just resorting to the default params if the first two come back empty.

enahum:
@shazm conflicts needs to be resolved

josephbaylon (changes requested):
@shazm I couldn't get it to work with either generic client or API client (see images below)

Steps,
1. Run `node retry_server.js`
2. See details below

|Generic Client|API Client|
|:---|:---|
|Screen Shot 2021-04-01 at 8 54 58 AM|Screen Shot 2021-04-01 at 8 57 16 AM|
|Screen Shot 2021-04-01 at 8 55 07 AM|Screen Shot 2021-04-01 at 8 58 06 AM|

Can you please try locally and see if you have the same results?

josephbaylon:
When I look at the server logs, it's even connecting to it.

enahum:
I don't get why @josephbaylon always have issues with the Android PR's.. @josephbaylon perhaps share expectations with @shazm, things like what should happen before assigning you as a reviewer... I believe going back and forth like this is causing too much time waste

josephbaylon:
I think generally for any PR I would expect the dev to have tested it first locally, provide the scope of what needs to be tested, and the steps on how to test the app properly. Then I would verify that it is indeed working (or not working) as a separate user from the dev (because usually it works on dev's environment but won't work on somebody else's or vice versa).

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -21,16 +21,19 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
 
     @ReactMethod
     fun createClientFor(baseUrl: String, options: ReadableMap, promise: Promise) {
+
+        // Don't trust user input...
+        val url = baseUrl.trimSlashes();
```
migbot, line 26: Since `.trimSlashes` is used in a few places, and to better explain the need for it, can this be placed in a `sanitizeUrl` func and comment it with something like "baseUrl is used as a key in various maps so we want to be sure it has a standard format"?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -40,7 +43,7 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     @ReactMethod
     fun getClientHeadersFor(baseUrl: String, promise: Promise) {
         try {
-            promise.resolve(sessionsRequest[baseUrl]?.build()?.headers?.readableMap())
+            promise.resolve(sessionsRequest[baseUrl.trimSlashes()]?.build()?.headers?.readableMap())
```
migbot, line 46: Like above, can you do `val url = baseUrl.trimSlashes();` and then use `url` here? I think it will be slightly more readable.

migbot: Same for all instances below.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -80,9 +86,12 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
 
     @ReactMethod
     fun post(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise) {
+        // Don't trust user input...
+        val url = baseUrl.trimSlashes();
+
         try {
-            val request = sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).post(options.bodyToRequestBody()).parseOptions(options, sessionsClient[baseUrl]!!, baseUrl).build();
-            sessionsClient[baseUrl]!!.build().newCall(request).execute().use { response ->
+            val request = sessionsRequest[url]!!.url(formUrlString(url, endpoint)).post(options.bodyToRequestBody()).parseOptions(options, sessionsClient[url]!!, url).build();
```
migbot, line 94: For readability, can you add newlines before/after each `.`?

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -212,6 +212,13 @@ fun formUrlString(baseUrl: String, endpoint: String): String {
     return baseUrl.toHttpUrlOrNull()!!.newBuilder().addPathSegments(endpoint.trim { c -> c == '/' }).build().toString()
 }
 
+/**
+ *
+ */
+fun String.trimSlashes(): String {
+    return this.trim { c -> c == '/' }
```
migbot, line 219: If this is only being used for the `baseUrl`, maybe it can just be done in the `sanitizeUrl` suggested above?

android/src/main/java/com/mattermost/networkclient/interceptors/RetryInterceptor.kt
```diff
@@ -36,22 +36,22 @@ class RetryInterceptor() : Interceptor {
         val request = chain.request()
         var response = chain.proceed(request)
         var attempts = 0;
-        val config = getRetryConfig(request.url.host)
+        val config = getRetryConfig(request.url.scheme + "://" + request.url.host + ":" + request.url.port )
```
migbot, line 39: If there is no port will the trailing `:` cause issues?

shazm:
Manual tests I did:

- Generic Client per-request linear and exponential retries
- API Client default config linear + exponential retries
- API Client "client" config linear + exponential retries
- API Client per-request config linear + exponential retries

And then I also tested with one of the existing clients, I think the RequestBin one to make sure that it hasn't impacted the standard / first attempt-success requests.

I'll see if I can replicate - android seems to be exceptionally sensitive to environment params (simulator, node, npm versions etc)

enahum:
> android seems to be exceptionally sensitive to environment params (simulator, node, npm versions etc)

This is a huge worry, try and mitigate this issue.. never seen a library that is as sensitive as this one when it comes to environment, please do harden it

shazm:
@josephbaylon I think the issue might be around the server; I haven't been able to figure it out -

I've added a `retry-lite.js` server which has a hard limit of 10 and have confirmed the following on it:

- Generic Client Linear + Exponential Retries
- APIClient Linear + Exponential Retries

These were with against `http://10.0.2.2:8000`, with retryLimit set to 2, 7, 15. For linear retries, I tested the interval of 500, 1500, and for exponential I tested 2/0.5 and 4/0.8. All worked as expected.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: In the methods where I call `...baseUrl.trimSlashes()` just once, I've called it directly - but if in that method I called `baseUrl` more than once, I moved it out into a new constant.

I'm not sure it's worthwhile making `url = baseUrl.trimSlashes()` on all instances.

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
shazm: I've been wondering about this - I think we might have to revisit the comment from before about using specific parts of the URL strictly; something like `scheme://host:port` - I think a `baseUrl` is too open to interpretation; so when a the `baseUrl` is passed in, we parse it into it's URL fragments and use them; this way there's no ambiguity around ports/trailing slashes etc.

The `trimSlashes()` was because of bad input; we really should be strict with input rather than try to capture and handle edge cases...

android/src/main/java/com/mattermost/networkclient/interceptors/RetryInterceptor.kt
shazm: Likely, see https://github.com/mattermost/react-native-network-client/pull/53#discussion_r607469209

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
shazm: How about something like https://github.com/mattermost/react-native-network-client/pull/53/commits/06404391a7d03e7f76ad2d9d89631fcf21f48a78 - it seems to work pretty well and sets a good default.

enahum: The problem about being strict with scheme://host:port is that you are leaving out the possibility of using a virtual folder or subpath as part of your baseUrl (scheme://host:port/path) for example https://shaz.com/mattermost

josephbaylon (approved):
LGTM. The retries now work as expected.

josephbaylon:
@shazm Just a note. I'll eventually remove `retry_server.js` once you merge this PR to master and I pull your changes to PR-57. I have modified the `mockserver.js` in PR-57 to have the retry capability. But I'll keep the new retry-lite.js for your own testing.

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -202,3 +226,10 @@ fun formUrlString(baseUrl: String, endpoint: String): String {
 fun emitEvent(reactContext: ReactContext, eventName: String, params: Any) {
     reactContext.getJSModule(RCTDeviceEventEmitter::class.java).emit(eventName, params)
 }
+
+/**
+ * Returns a "Base URL" for a given HTTP URL
+ */
+fun HttpUrl.toBaseUrl(): String {
+    return "${this.scheme}://${this.host}:${this.port}"
```
migbot, line 234: This will be a problem when the "base url" contains a subpath as mentioned by @enahum. How about just doing `.toHttpUrl().toString()`?

josephbaylon:
@shazm For parity, iOS displays the responses to response overlay for both failure and success attempts (see images below). Currently, I believe Android shows failure attempts on error modal.

Please see https://github.com/mattermost/react-native-network-client/pull/58 .

Please apply the same for both **linear** and **exponential** scenarios,

|iOS Failure Attempt|iOS Successful Attempt|
|:---|:---|
|Screen Shot 2021-04-06 at 2 39 50 PM|Screen Shot 2021-04-06 at 2 40 12 PM|

enahum:
@shazm were are we on this?

shazm:
Just pushed up an update; this now has:

- HTTP Status Codes that should be retried; defaults to Alamo Fire defaults (408, 500, 502, 503, 504)
- HTTP Methods that should be retried; defaults to GET/POST/PUT/PATCH/DELETE
- Example app allows input of HTTP Status Codes, and selecting methods

There's a few other changes worth noting:

- All HTTP responses are now resolved as a successful promise, any other type of response is thrown as an exception

@types/APIClients.ts
```diff
@@ -103,6 +103,8 @@ type RetryPolicyConfiguration = {
     retryInterval?: number;
     exponentialBackoffBase?: number;
     exponentialBackoffScale?: number;
+    statusCodes?: number[];
```
migbot, line 106: For consistency with interval and methods:
```suggestion
    retryStatusCodes?: number[];
```

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -5,30 +5,36 @@ import com.facebook.react.bridge.*
...
+    private val sessionsConfig = SessionsObject.config
 
     override fun getName(): String {
         return "APIClient"
     }
 
     @ReactMethod
     fun createClientFor(baseUrl: String, options: ReadableMap, promise: Promise) {
+        // Don't trust user input...
+        val url = baseUrl.toHttpUrl().toString();
```
migbot, line 27: Something about having this same comment in multiple places leads me to want to either:
1. Remove the comment or
2. Create a function that returns `url.toHttpUrl().toString()` and comment that function only

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
migbot: ^^^ here and below

android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -3,56 +3,61 @@ package com.mattermost.networkclient
...
+        client.addInterceptor(RetryInterceptor("generic"))
+    }
+
     @ReactMethod
     fun get(url: String, options: ReadableMap, promise: Promise) {
-        val request = Request.Builder().url(url).parseOptions(options, client).build();
+        val request = Request.Builder().url(url).parseOptions(options, client, url).build();
         client.build().newCall(request).execute().use { response ->
-            response.promiseResolution(promise)
+            promise.resolve(response.returnAsWriteableMap("generic"))
```
migbot, line 28: What's "generic" used for?

example/src/components/RetryPolicyConfiguration.tsx
```diff
@@ -27,17 +37,26 @@ const RetryPolicyConfiguration = (props: RetryPolicyConfigurationProps) => {
     const onExponentialPress = () =>
         onCheckBoxPress(Constants.RETRY_TYPES.EXPONENTIAL_RETRY);
     const onCheckBoxPress = (policyType: RetryTypes) => {
-        if (policyType === props.policyType) {
-            props.onTypeSelected(undefined);
-        } else {
-            props.onTypeSelected(policyType);
-        }
+        props.onTypeSelected(policyType);
```
migbot, line 40: We should be able to uncheck and leave retry type undefined, no?

example/src/components/RetryPolicyConfiguration.tsx
```diff
@@ -1,24 +1,34 @@
...
     policyType?: RetryTypes;
-    retryLimit?: number;
-    retryInterval?: number;
-    exponentialBackoffBase?: number;
-    exponentialBackoffScale?: number;
+    // retryLimit?: number;
+    // retryInterval?: number;
+    // exponentialBackoffBase?: number;
+    // exponentialBackoffScale?: number;
+    // statusCodes?: number[];
```
migbot, line 24: Why are these removed?

example/src/hooks/index.tsx
```diff
@@ -94,25 +94,35 @@ type UseRetryPolicyConfigurationResponse = [
     (retryLimit: number) => void,
     (retryInterval: number) => void,
     (exponentialBackoffBase: number) => void,
-    (exponentialBackoffScale: number) => void
+    (exponentialBackoffScale: number) => void,
+    (statusCodes: number[]) => void,
+    (retryMethods: string[]) => void
 ];
 
+// enum RetryTypes =
```
migbot, line 102:
```suggestion
```

ios/NetworkClient.swift
```diff
@@ -135,20 +135,27 @@ extension NetworkClient {
...
-            let retryPolicy = LinearRetryPolicy(retryLimit: retryLimit, retryInterval: retryInterval)
+            let retryPolicy = LinearRetryPolicy(retryLimit: retryLimit, retryInterval: retryInterval, retryableHTTPStatusCodes: retryStatusCodes)
             retriers.append(retryPolicy)
         } else if configuration["type"].string == RETRY_TYPES["EXPONENTIAL_RETRY"] {
             let retryLimit = configuration["retryLimit"].uInt ?? ExponentialRetryPolicy.defaultRetryLimit
             let exponentialBackoffBase = configuration["exponentialBackoffBase"].uInt ?? ExponentialRetryPolicy.defaultExponentialBackoffBase
             let exponentialBackoffScale = configuration["exponentialBackoffScale"].double ?? ExponentialRetryPolicy.defaultExponentialBackoffScale
 
-            let retryPolicy = ExponentialRetryPolicy(retryLimit: retryLimit, exponentialBackoffBase: exponentialBackoffBase, exponentialBackoffScale: exponentialBackoffScale)
+            let retryPolicy = ExponentialRetryPolicy(retryLimit: retryLimit, exponentialBackoffBase: exponentialBackoffBase, exponentialBackoffScale: exponentialBackoffScale, retryableHTTPStatusCodes: retryStatusCodes)
```
migbot, line 158: 💯 , just gotta do the same for HTTP methods.

PR #55, opened 2021-03-19
MM-34376 Android Certificates

shazm:
#### Summary
This PR Introduces handling of the bearerAuthTokenHeader, and attaching client certificates to the client for TLS.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34376

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
+++ b/android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
@@ -2,12 +2,18 @@ package com.mattermost.networkclient.helpers
 
 import com.facebook.react.bridge.*
 import com.facebook.react.modules.core.DeviceEventManagerModule.RCTDeviceEventEmitter
-import okhttp3.*
-import java.util.concurrent.TimeUnit
 import com.mattermost.networkclient.interceptors.*
+import okhttp3.*
 import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.tls.HandshakeCertificates
 import org.json.JSONObject
+import java.io.FileInputStream
+import java.security.KeyStore
+import java.security.cert.X509Certificate
+import java.util.*
+import java.util.concurrent.TimeUnit
+import javax.net.ssl.KeyManagerFactory
 
 /**
  * Parses the response data into the format expected by the App
@@ -73,18 +79,12 @@ fun Request.Builder.parseOptions(options: ReadableMap?, session: OkHttpClient.Bu
 fun OkHttpClient.Builder.parseOptions(options: ReadableMap?, request: Request.Builder?): OkHttpClient.Builder {
     if(options == null) return this;
 
-    // Following Redirects
-    if (options.hasKey("followRedirects")) {
-        val followRedirects = options.getBoolean("followRedirects")
-        this.followRedirects(followRedirects)
-        this.followSslRedirects(followRedirects)
-    }
-
-    // Retries
+    // Retries off by default
     this.retryOnConnectionFailure(false);
+
+    // Set retry policy if provided
     if (options.hasKey("retryPolicyConfiguration")) {
         val retryPolicyConfiguration = options.getMap("retryPolicyConfiguration")!!.toHashMap();
-
         val retryType = retryPolicyConfiguration["type"] as String?
         val retryLimit = retryPolicyConfiguration["retryLimit"] as Double?
         val retryExponentialBackoffBase = retryPolicyConfiguration["exponentialBackoffBase"] as Double?
@@ -135,6 +135,51 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap?, request: Request.Bu
         if (sessionConfig.hasKey("enableCompression")) {
             this.minWebSocketMessageToCompress(0);
         }
+    }
+
+    // Request Adapter Config
+    if(options.hasKey("requestAdapterConfiguration")){
+        val requestAdapterConfiguration = options.getMap("requestAdapterConfiguration")!!;
+        if(requestAdapterConfiguration.hasKey("bearerAuthTokenResponseHeader")){
+            val bearerAuthTokenResponseHeader = requestAdapterConfiguration.getString("bearerAuthTokenResponseHeader")!!
+            this.addInterceptor(BearerTokenInterceptor(bearerAuthTokenResponseHeader))
+        }
+    }
+
+    // Client Cert Config
+    if(options.hasKey("clientP12Configuration")){
+        val clientP12Configuration = options.getMap("clientP12Configuration")!!;
+
+        // Get our Cert
+        val path = clientP12Configuration.getString("path")!!;
+        val password = if(clientP12Configuration.hasKey("password")) {
+            clientP12Configuration.getString("password")!!
+        } else {
+            ""
+        }
+
+        // Save P12 into KeyStore
+        val aliases = getAndSaveCertsFromP12(path, password)
+
+        // Use it for our Client
+        val keyStore = KeyStore.getInstance("AndroidKeyStore")
+        keyStore.load(null, null)
+
+        // Certificates Builder
+        val certificatesBuilder = HandshakeCertificates
+                .Builder()
+                .addPlatformTrustedCertificates()
+
+        // Get all the saved certificates from KeyStore
+        while(aliases.hasMoreElements()){
+            val alias = aliases.nextElement() as String
+            val cert: X509Certificate = keyStore.getCertificate(alias) as X509Certificate
+            certificatesBuilder.addTrustedCertificate(cert)
+        }
+
+        // Build our certificates and attach them to the client
+        val certificates = certificatesBuilder.build()
+        this.sslSocketFactory(certificates.sslSocketFactory(), certificates.trustManager)
     }
 
     return this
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -73,18 +79,12 @@ fun Request.Builder.parseOptions(options: ReadableMap?, session: OkHttpClient.Bu
 fun OkHttpClient.Builder.parseOptions(options: ReadableMap?, request: Request.Builder?): OkHttpClient.Builder {
     if(options == null) return this;
 
-    // Following Redirects
-    if (options.hasKey("followRedirects")) {
```
enahum, line 77: Why is this being removed?

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -73,18 +79,12 @@ fun Request.Builder.parseOptions(options: ReadableMap?, session: OkHttpClient.Bu
...
-        this.followRedirects(followRedirects)
-        this.followSslRedirects(followRedirects)
-    }
-
-    // Retries
+    // Retries off by default
     this.retryOnConnectionFailure(false);
+
+    // Set retry policy if provided
     if (options.hasKey("retryPolicyConfiguration")) {
```
enahum, line 85: Should you add the linear retry in this PR or is it in a separate one?

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -137,6 +137,51 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap?, request: Request.Bu
...
+        } else {
+            ""
+        }
+
+        // Save P12 into KeyStore
+        val aliases = getAndSaveCertsFromP12(path, password)
+
+        // Use it for our Client
+        val keyStore = KeyStore.getInstance("AndroidKeyStore")
+        keyStore.load(null, null)
```
enahum, line 166: What does this do?

enahum (commented):
A few comments but the more important one is this:

Should the alias be the baseUrl?

This is how I see this thing working.
1. Configure the client without the cert path
2. Execute a request
3. Get the error that a cert is needed
4. Load a cert into the app
5. Configure the Client to use the certificate which Stores the cert in the keystore
6. Execute requests successfully
----
1. Open the app
2. Configure the client without the cert
3. Lib detects that there is a cert already stored with this baseUrl in the keystore and applies it
4. Execute requests successfully
----
1. A logout is issued
2. Invalidate client and remove/delete cert from keystore

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
shazm: It exists in `SessionConfiguration` - this is just cleaning up from an earlier change.

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
shazm: I have it here separately already: https://github.com/mattermost/react-native-network-client/pull/53

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
shazm: It just "loads" the keystore; https://developer.android.com/training/articles/keystore#WorkingWithKeyStoreEntries

shazm:
> A few comments but the more important one is this:
> 
> Should the alias be the baseUrl?
> 
> This is how I see this thing working.
> 
> 1. Configure the client without the cert path
> 2. Execute a request
> 3. Get the error that a cert is needed
> 4. Load a cert into the app
> 5. Configure the Client to use the certificate which Stores the cert in the keystore
> 6. Execute requests successfully
> 
> 1. Open the app
> 2. Configure the client without the cert
> 3. Lib detects that there is a cert already stored with this baseUrl in the keystore and applies it
> 4. Execute requests successfully
> 
> 1. A logout is issued
> 2. Invalidate client and remove/delete cert from keystore

I was actually thinking about this; and I actually had set the alias as the base URL - but if the PK12 has multiple certificates that get loaded into the keystore - I *think* having the same alias might cause issues when retrieving / saving them. I can test it though.

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -199,3 +244,40 @@ fun formUrlString(baseUrl: String, endpoint: String): String{
...
+    try {
+        // Get an instance of the AndroidKeyStore
+        val androidKeyStore: KeyStore = KeyStore.getInstance("AndroidKeyStore")
+        androidKeyStore.load(null)
+
+        // Create an instance of a pkcs12 keystore
+        val keyStore: KeyStore = KeyStore.getInstance("pkcs12")
+
+        // Load our certificate file into it, with the password
+        keyStore.load(FileInputStream(pathToFile), password.toCharArray())
```
migbot, line 265: This will throw if the password is wrong?

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -137,6 +137,51 @@ fun OkHttpClient.Builder.parseOptions(options: ReadableMap?, request: Request.Bu
...
+        // Get all the saved certificates from KeyStore
+        while(aliases.hasMoreElements()){
+            val alias = aliases.nextElement() as String
+            val cert: X509Certificate = keyStore.getCertificate(alias) as X509Certificate
+            certificatesBuilder.addTrustedCertificate(cert)
+        }
+
+        // Build our certificates and attach them to the client
+        val certificates = certificatesBuilder.build()
+        this.sslSocketFactory(certificates.sslSocketFactory(), certificates.trustManager)
```
migbot, line 182: So adding them this way doesn't necessitate us manually grabbing the cert from the Keychain and providing it during the handshake?

android/src/main/java/com/mattermost/networkclient/interceptors/BearerTokenInterceptor.kt
```diff
@@ -0,0 +1,33 @@
...
+class BearerTokenInterceptor(private val bearerAuthTokenResponseHeader: String) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
+        var request = chain.request()
+        var token = SessionsObject.data[request.url.host]?.get("AuthToken")
+
+        // Check if there is a token already, if so - add it
+        if(token !== null){
+            token = SessionsObject.data[request.url.host]?.get("AuthToken") as String?
```
migbot, line 17: Why do you need to get it again?

android/src/main/java/com/mattermost/networkclient/interceptors/BearerTokenInterceptor.kt
```diff
@@ -0,0 +1,33 @@
...
+            request = request.newBuilder().addHeader("Authorization", "Bearer $token").build()
+        }
+
+        // Run our request
+        val response = chain.proceed(request)
+
+        // Save the token for future requests
+        token = response.header("bearerAuthTokenResponseHeader")
```
migbot, line 25: Shouldn't this be?
```suggestion
        token = response.header(bearerAuthTokenResponseHeader)
```

josephbaylon:
@shazm Please take note of this current limitation in iOS and apply a fix to android https://github.com/mattermost/react-native-network-client/pull/57#issuecomment-809725355 Let me know if you need me to create a separate ticket.

shazm:
> @shazm Please take note of this current limitation in iOS and apply a fix to android https://github.com/mattermost/react-native-network-client/pull/57#issuecomment-809725355 Let me know if you need me to create a separate ticket.

Will check, but shouldn't need a separate ticket - I'm thinking about rebase this work off the E2E tests though and see if I can get close to getting the same tests passing for Android - it should make merging easier once both are ready.

josephbaylon:
@shazm Thanks. Here's another one that should also need to work in Android https://github.com/mattermost/react-native-network-client/pull/57#issuecomment-809736300

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
enahum: Do we need to load it for every client that gets built? Does this have any perf implication?

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
enahum: Also interested in the answer to this

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -199,3 +244,40 @@ fun formUrlString(baseUrl: String, endpoint: String): String{
...
+        // Get all the key aliases
+        aliases = keyStore.aliases()
+
+        // Loop through them
+        while (aliases.hasMoreElements()) {
+            val alias = aliases.nextElement() as String
+            val cert: X509Certificate = keyStore.getCertificate(alias) as X509Certificate
+
+            // Add the certificate to the KeyStore
+            androidKeyStore.setCertificateEntry(alias, cert)
```
enahum, line 276: Do we use any flags to save/retrieve from the store?

Like: should the device be unlocked? Or can this be access at all times? I guess no hardware cryptography is needed, so for software do you know which one is used?

@migbot Same question I guess applies to iOS and the keychain

android/src/main/java/com/mattermost/networkclient/interceptors/BearerTokenInterceptor.kt
enahum: Same question

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
migbot: @enahum I used [kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly) for accessing both token and cert from the Keychain.

enahum: > @enahum I used [kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly) for accessing both token and cert from the Keychain.

So backup and restore across devices will require us to import the cert again, yes?

And things like background tasks or fetching notification content + ack will only be possible after the first unlock.

This one allows backup https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlock

Obviously always allowing access is now deprecated, so I think we should document this, an be opened to someone contributing this value as a setting

shazm:
Refs:

https://developer.android.com/training/articles/keystore
https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/CustomTrust.java
https://gist.github.com/handstandsam/d7d7956a05698b23cad14dd404350e23

PR #57, opened 2021-03-25
[MM-33806, MM-34097] Detox/E2E: Sample iOS E2E for certificate auth

josephbaylon:
#### Summary
- Added e2e for secure api client (`get/put/patch/post/delete`), file client (`upload/fast image`),  and websocket client
- Added new detox screen for APIClientImportP12
- Removed third-party mockserver; created in-house mockserver
- Added secure capability to mockserver, file server, and websocket server
- Added ability to download binary file from url and write binary file to native
- Added new testIDs and helper functions

#### Ticket Link
JIRA:
- https://mattermost.atlassian.net/browse/MM-33806
- https://mattermost.atlassian.net/browse/MM-34097

Fixes merged in this PR
- https://mattermost.atlassian.net/browse/MM-34420
- https://mattermost.atlassian.net/browse/MM-34374
- https://mattermost.atlassian.net/browse/MM-34218
- https://mattermost.atlassian.net/browse/MM-34612
- https://mattermost.atlassian.net/browse/MM-34422

Screen Shot 2021-04-08 at 9 00 38 AM

PR #58, opened 2021-03-31
[MM-34374, MM-34420, MM-34612] Various fixes

migbot:
#### Summary
* Adds `trustSelfSignedServerCertificate` to the WebSocketClient configuration.
* Uses `host` as the `kSecAttrLabel` for identity/certificate Keychain items. This allows both the APIClient and WebSocketClient to retrieve the correct certificate when their `host` values match. If a P12 has already been imported, further import attempts will trigger a client error event but the certificate will be found for the handshake and requests will be authenticated.
* Registered the client error callback prior to making the native function call so that any error events will be handled.
* Throw error on P12 import for the WebSocket client. This prevents the WebSocket client from being created if an error other than a duplicate identity is encountered.
* Reject promise if error is encountered when calling `importClientP12` on an APIClient.
* Resolve promise for `requestRetryFailed` errors if underlying error has a `responseCode` and also include `retriesExhausted: true` in response

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34374
https://mattermost.atlassian.net/browse/MM-34420
https://mattermost.atlassian.net/browse/MM-34612


josephbaylon (changes requested):
@migbot I'm getting the following when trying to build ios example app

Screen Shot 2021-03-30 at 11 48 09 PM

ios/WebSocketManager.swift
```diff
@@ -30,6 +30,7 @@ class WebSocketManager: NSObject {
         var request = URLRequest(url: url)
         var compressionHandler: CompressionHandler? = nil
         var clientCredential: URLCredential? = nil
+        var certPinner: FoundationSecurity = nil
```
migbot, line 33:
```suggestion
        var certPinner: FoundationSecurity? = nil
```

josephbaylon:
Changes copied to https://github.com/mattermost/react-native-network-client/pull/57 and should be testable there

josephbaylon:
@migbot For last commit, please revert the rejection for `APIClient.swift` so tests can pass, only websocket should have the rejection

enahum:
> > @migbot For last commit, please revert the rejection for `APIClient.swift` so tests can pass, only websocket should have the rejection
> 
> 
> 
> Hey Joseph, as mentioned in our chat, there are two cases in which the APIClient can import a P12. The first is when creating the APIClient. In this case we do not reject the promise as we don't want to prevent creation of the client if there's an error importing the P12 (ie, incorrect password). In this case the native side just sends an error event. Once the APIClient is created, the second case, client function `importClientP12`, is available for the client to retry the import. Since this function deals solely with importing the P12, we reject here if there is an error.
> 
> 
> 
> For the WebSocketClient, there's a limitation in that creating the client is coupled with P12 import since it needs to be passed down to the underlying Starscream library. For this reason we 1) don't provide a separate `importClientP12` function and 2) we reject if there is an import error during creation of the client.

I've got lost here... this is what I would expect

1. Create an API client
2. Attach or import a cert
3. Create a WS client
4. Connect the WS -> this uses the cert imported in step 2

Does that work?

josephbaylon:
@enahum The WS client is working fine and the restriction to not be able to import P12 once the WS client is created is in place. The rejection on WebSocketClient is then reasonable since there's no way for user to import new P12.

The issue is on the API client where we're allowing import of p12 once client has been created.

We have a `clientErrorEventHandler` that can be enabled/disabled in the preset API clients but always enabled when creating API client from the UI.

@migbot's concern, I believe, was that when p12 is invalid or has incorrect password that it wouldn't prompt the user with client error, that's why he added the rejection on the `APIClient`. But it does on both API client creation and when import to existing API client.

My original suggestion is to only apply the rejection on WebSocketClient but NOT to APIClient as doing it on APIClient would cause unnecessary promise rejection on cases where p12 has already been imported BUT the `clientErrorEventHandler` has been disabled, for reasons that we're not testing the error in those cases and the tests need to be idempotent.

We will add tests around the actual errors later when `clientErrorEventHandler` is enabled, but that's a separate concern.

josephbaylon:
Thus, I think the change in `APIClientImportP12Screen` is unnecessary and the rejection on `APIClient` needs to be reverted

josephbaylon:
@migbot Thanks for the update. We need to remove the navigate.back() so that whether or not the user receives an error, he's not automatically navigated back in terms of tests that don't care if the import was successful or not

josephbaylon (approved):
LGTM. I applied the changes in pr-57 and works there. However, I had to remove the navigate.back() in my PR to adjust to workflow for either successful/failed import

josephbaylon (approved):

josephbaylon:
@migbot I'll re-approve once you've pushed the fixes for retry and error codes. Thanks

ios/WebSocketManager.swift
```diff
@@ -46,23 +47,29 @@ class WebSocketManager: NSObject {
...
                 let path = clientP12Configuration["path"]
                 let password = clientP12Configuration["password"]
+                
                 do {
-                    try Keychain.importClientP12(withPath: path!, withPassword: password, forServerUrl: url.absoluteString)
-                    let (identity, certificate) = try Keychain.getClientIdentityAndCertificate(for: url.absoluteString)!
-                    clientCredential = URLCredential(identity: identity, certificates: [certificate], persistence: URLCredential.Persistence.permanent)
+                    try Keychain.importClientP12(withPath: path!, withPassword: password, forHost: url.host!)
+                } catch KeychainError.DuplicateIdentity {
+                    // do nothing
```
enahum, line 54: 💯

enahum (approved):

enahum (approved):

josephbaylon (approved):
LGTM. Copied changes to PR-57 and tested and linear and exponential retry works and failure/success response goes to response overlay.

PR #65, opened 2021-04-16
[MM-30124, MM-34747]Android - Refactor, fix linear retry, and bearer auth interceptor

shazm:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30124
https://mattermost.atlassian.net/browse/MM-34747


example/retry_server.js
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
josephbaylon, line 1: We can remove this file as the retry mechanism is already incorporated in `servers/mockserver.js`

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -56,33 +75,56 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+            promise.resolve(sessionsClient.keys);
         } catch (err: Throwable) {
             promise.reject(err)
         }
     }
 
     @ReactMethod
     fun get(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise) {
+        // Don't trust user input...
+        val url = baseUrl.toHttpUrl().toString();
```
avinashlng1080, line 93: Can baseUrl be null ? Like have we tested with 'wrong data' ?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -158,26 +228,26 @@ class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
-                    "PATCH" -> sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).patch(body)
+                    "PUT" -> Request.Builder().url(formUrlString(url, endpoint)).put(body)
+                    "PATCH" -> Request.Builder().url(formUrlString(url, endpoint)).patch(body)
                     // Default to "POST"
-                    else -> sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).post(body)
+                    else -> Request.Builder().url(formUrlString(url, endpoint)).post(body)
                 }
             } else {
-                sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).post(body)
+                Request.Builder().url(formUrlString(url, endpoint)).post(body)
```
avinashlng1080, line 239: Might be a silly comment coming from outside but do we need GET method ?  I see we have put, patch and post but no get.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: I think @migbot has a change coming up soon that wraps .toHttpUrl in a try catch

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
shazm: I don't think you're supposed to send body data with a GET request; might be wrong - @enahum or @migbot thoughts?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
migbot: Made the change in https://github.com/mattermost/react-native-network-client/pull/65/commits/34930a0b68c6c25b1853d063ed8f9a002947a9e4

android/src/main/java/com/mattermost/networkclient/helpers/Helper.kt
```diff
@@ -12,6 +12,14 @@ import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.RequestBody.Companion.toRequestBody
 import org.json.JSONObject
 
+var Response.retriesExhausted: Boolean
```
migbot, line 15: @shazm Made this change to avoid storing `retriesExhausted` in `SessionsObject`.

migbot: Disregard, this didn't work out :/

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
migbot: I read that same response and [allowed it on iOS](https://github.com/mattermost/react-native-network-client/blob/master/ios/NetworkClient.swift#L59) for simplicity in the native API.

@types/APIClients.ts
```diff
@@ -103,6 +103,8 @@ type RetryPolicyConfiguration = {
     retryInterval?: number;
     exponentialBackoffBase?: number;
     exponentialBackoffScale?: number;
+    statusCodes?: number[];
+    retryMethods?: string[];
```
shazm, line 107: Since we're tidying this - worth getting the naming here consistent? So everything is either pretended with `retry` or nothing is?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -1,186 +1,146 @@
...
+import java.io.IOException
+import java.lang.Exception
 import kotlin.collections.HashMap
 
 class APIClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
-
-    var sessionsClient = mutableMapOf<String, OkHttpClient.Builder>()
-    var sessionsRequest = mutableMapOf<String, Request.Builder>()
-    var calls = mutableMapOf<String, Call>()
+    private val clients = mutableMapOf<HttpUrl, NetworkClient>()
```
shazm, line 17: Something I didn't think of yesterday; I used the singleton pattern because the global interceptor needs an already initialised object where it can call the existing clients / interceptors / configs - I'm wondering if this will still work because NetworkClient is now a class that needs to be initialised in order to be used? Unless there's a way of passing around the NetworkClient class that already exists to the parent app...

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -1,186 +1,146 @@
...
     fun createClientFor(baseUrl: String, options: ReadableMap, promise: Promise) {
+        var url: HttpUrl
         try {
-            // Create the client and request builder
-            sessionsClient[baseUrl] = OkHttpClient().newBuilder();
-            sessionsRequest[baseUrl] = Request.Builder().url(baseUrl);
-
-            // Attach client options if they are passed in
-            sessionsClient[baseUrl]!!.parseOptions(options, sessionsRequest[baseUrl]);
+            url = baseUrl.toHttpUrl()
```
shazm, line 28: 🙌

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -1,186 +1,146 @@
...
-            promise.resolve(sessionsRequest[baseUrl]?.build()?.headers?.readableMap())
-        } catch (error: Error) {
+            url = baseUrl.toHttpUrl()
+        } catch (error: IllegalArgumentException) {
+            return promise.reject(error)
+        }
+
+        val map = Arguments.createMap()
+        val headers = clients[url]!!.clientHeaders
+        for((k, v) in headers.toHashMap()){
```
shazm, line 52: I think this is used in a few places, might be worth making a little helper?

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
```diff
@@ -1,186 +1,146 @@
...
     fun get(baseUrl: String, endpoint: String, options: ReadableMap, promise: Promise) {
-        try {
-            val request = sessionsRequest[baseUrl]!!.url(formUrlString(baseUrl, endpoint)).parseOptions(options, sessionsClient[baseUrl]!!).build();
-            sessionsClient[baseUrl]!!.build().newCall(request).execute().use { response ->
-                response.promiseResolution(promise)
-            }
-        } catch (e: IOException) {
-            promise.reject(e)
-        }
+        request("GET", baseUrl, endpoint, options, promise)
```
shazm, line 99: So clean! 🙌

android/src/main/java/com/mattermost/networkclient/Extensions.kt
```diff
@@ -0,0 +1,66 @@
+package com.mattermost.networkclient
+
```
shazm, line 2: Doh good idea, didn't even think of holding the extensions in their own file - makes a lot more sense 👍

android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -1,59 +1,48 @@
...
-import com.mattermost.networkclient.helpers.promiseResolution
-import okhttp3.OkHttpClient
-import okhttp3.Request
-
+import java.lang.Exception
 
 class GenericClientModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
-
-    var client = OkHttpClient().newBuilder();
+    private var client = NetworkClient();
```
shazm, line 7: 🙌

android/src/main/java/com/mattermost/networkclient/GenericClientModule.kt
```diff
@@ -1,59 +1,48 @@
...
         return "GenericClient"
     }
 
     @ReactMethod
     fun get(url: String, options: ReadableMap, promise: Promise) {
-        val request = Request.Builder().url(url).parseOptions(options, client).build();
-        client.build().newCall(request).execute().use { response ->
-            response.promiseResolution(promise)
-        }
+        request("GET", url, options, promise)
```
shazm, line 15: SO CLEAN! 🙌

example/src/components/RetryPolicyConfiguration.tsx
```diff
@@ -122,6 +157,52 @@ const RetryPolicyConfiguration = (props: RetryPolicyConfigurationProps) => {
...
+                                containerStyle={{ height: 50 }}
+                                inputContainerStyle={{
+                                    borderColor: "rgba(255,255,255,0)",
+                                }}
+                            />
+
+                            <ButtonGroup
+                                selectedIndexes={selectedMethodsIndex}
+                                // @ts-ignore
+                                onPress={setSelectedMethodsIndex}
```
shazm, line 198: I think it might be worth pushing a PR request further upstream to fix the types, I did put in the ts-ignore but it always leads to bad things!

example/src/hooks/index.tsx
```diff
@@ -23,8 +23,8 @@ export const useSessionConfiguration = (): UseSessionConfigurationResponse => {
         followRedirects: true,
         allowsCellularAccess: true,
         waitsForConnectivity: false,
-        timeoutIntervalForRequest: 30,
-        timeoutIntervalForResource: 30,
+        timeoutIntervalForRequest: 30000,
```
shazm, line 26: All time related configs are now milliseconds? (Just double checking!)

shazm (commented):
Looks pretty awesome, just a few comments and questions.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
migbot: Ah, OK I'll look into this today!

example/src/hooks/index.tsx
migbot: For retries, yep. iOS uses seconds and Android milliseconds so decided to use milliseconds everywhere.

josephbaylon:
@migbot Please let me know which ones of the outstanding issues you're going to address in this PR; otherwise I can go ahead and approve this PR.

@types/APIClients.ts
migbot: Yep, will use some standard naming (prob default to AF names) in a separate PR.

android/src/main/java/com/mattermost/networkclient/APIClientModule.kt
migbot: Yep, will do so in a separate PR.

example/src/components/RetryPolicyConfiguration.tsx
migbot: For now will leave as is, since it's just our example app.

enahum (approved):

josephbaylon (approved):
LGTM. Re-tested linear retry and bearer auth interceptor and works fine.

PR #66, opened 2021-04-22
Android refactor

migbot:
#### Summary
In keeping somewhat inline with the iOS implementation, I moved things around and created a NetworkClient class that acts as a proxy to OkHttpClient.


enahum (approved):
LGTM.. @shazm perhaps you can be a bit more meticulous than me given your knowledge here.

PR #70, opened 2021-04-29
Android: Store bearer token from response, retrieven and apply for requests

migbot:
#### Summary
* Adds a KeyStoreHelper class to encrypt/decrypt the bearer token before storing / after retrieving from shared preferences.
* Adds an interceptor to retrieve the bearer token from a response's headers or apply it to a request's headers.
* Sets an expiration on cookies when "clearing" otherwise the cookie is still send with the request.

Note: To test this I manually cleared cookies for the requests in order for my MM local server not to receive them and instead use the token applied in the interceptor.


android/build.gradle
```diff
@@ -29,7 +29,7 @@ android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   buildToolsVersion getExtOrDefault('buildToolsVersion')
   defaultConfig {
-    minSdkVersion 21
+    minSdkVersion 24
```
enahum, line 32: Is this needed? Can we avoid it?

enahum (changes requested):

android/build.gradle
migbot: Reverted. I originally did this because I was playing with a library that required this.

enahum (approved):

android/build.gradle
migbot: Oh @enahum, actually the `KeyGenParameterSpec.Builder` code requires Android M. By removing this I need to build the spec differenty for < Android M. Let me give this a shot.

migbot: See https://stackoverflow.com/a/39264236

enahum (approved):

josephbaylon (approved):
LGTM

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    init {
+        setAndLoadKeyStore()
+        generateSecretKeyPairIfNeeded()
+    }
+
+    fun encryptData(data: String): String {
+        var temp = data
+        while (temp.toByteArray().size % 16 != 0) {
+            temp += "\u0020"
```
enahum, line 34: Hey @migbot do you remember why this was needed at the time?

migbot: I believe because of this: https://stackoverflow.com/a/46911659

android/build.gradle
enahum: Awesome thanks

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
enahum: awesome thanks

PR #71, opened 2021-04-29
[MM-30594, MM-34376] Android: Client certificate

migbot:
#### Handling of client certificate
Importing of the client certificate is done by:
1. Loading the user provided P12 file as a PKCS12 KeyStore
2. Looping through aliases
3. Grabbing only the first entry that is key entry
4. Loading a new PKCS12 KeyStore
5. Writing the key and certificate entries into this new KeyStore
6. Creating a file using `APIClientModule.context.openFileInput` and storing the KeyStore there using the user provided password
7. Encrypting the password and storing in SharedPreferences

#### Using the client certificate:
1. Retrieve and decrypt the password from SharedPreferences
2. Load our created PKCS12 KeyStore using the password
3. Retrieve the private key, client certificate, and intermediate certificates from the KeyStore and use them to create an OkHttp3 HeldCertificate which is then used to build a HandshakeCertificates
4. Use the HandshakeCertificates to add an SSLSocketFactory to the OkHttpClient

#### Handling trusting of self-signed server certificate
1. Added a `trustSelfSignedServerCertificate` instance property on NetworkClient
2. When this property is set to true:
* `addInsecureHost` for the client's baseUrl.host is added on HandshakeCertificates. This trusts any certificate presented by the server
* `hostnameVerifier` is set on the OkHttpClient to always return true. This means that we will not check that the server's certificate matches on hostname. This is required for our testing since our self-signed certificates are not created for real domains. I've added a note to the README file, but maybe we should consider not skipping hostname verification?

https://mattermost.atlassian.net/browse/MM-30594
https://mattermost.atlassian.net/browse/MM-34376

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/Extensions.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Extensions.kt
@@ -57,11 +57,11 @@ fun String.trimTrailingSlashes(): String {
 }
 
 /**
- * Computes the MD5 of a string
+ * Computes the SHA-256 hash of a string
  */
-fun String.md5(): String {
-    val md = MessageDigest.getInstance("MD5")
-    return BigInteger(1, md.digest(toByteArray()))
-            .toString(16)
-            .padStart(32, '0')
+fun String.sha256(): String {
+    return MessageDigest
+            .getInstance("SHA-256")
+            .digest(toByteArray())
+            .fold("", { str, it -> str + "%02x".format(it) })
 }
--- a/android/src/main/java/com/mattermost/networkclient/Extensions.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Extensions.kt
@@ -50,6 +50,13 @@ fun Request.Builder.applyHeaders(headers: ReadableMap?): Request.Builder {
 }
 
 /**
+ * Trims trailing slashes in the string
+ */
+fun String.trimTrailingSlashes(): String {
+    return trimEnd { c -> c == '/' }
+}
+
+/**
  * Computes the MD5 of a string
  */
 fun String.md5(): String {
--- a/android/src/main/java/com/mattermost/networkclient/Extensions.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Extensions.kt
@@ -5,6 +5,8 @@ import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableMap
 import okhttp3.Request
 import okhttp3.Response
+import java.math.BigInteger
+import java.security.MessageDigest
 
 var Response.retriesExhausted: Boolean? by NetworkClient.RequestRetriesExhausted
 
@@ -45,4 +47,14 @@ fun Request.Builder.applyHeaders(headers: ReadableMap?): Request.Builder {
     }
 
     return this;
+}
+
+/**
+ * Computes the MD5 of a string
+ */
+fun String.md5(): String {
+    val md = MessageDigest.getInstance("MD5")
+    return BigInteger(1, md.digest(toByteArray()))
+            .toString(16)
+            .padStart(32, '0')
 }
--- a/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
+++ b/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
@@ -3,14 +3,17 @@ package com.mattermost.networkclient
 import com.mattermost.networkclient.enums.RetryTypes
 import com.mattermost.networkclient.interfaces.RetryInterceptor
 import com.mattermost.networkclient.interceptors.*
+import com.mattermost.networkclient.helpers.KeyStoreHelper
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableMap
 import okhttp3.*
 import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.tls.HandshakeCertificates
 import org.json.JSONObject
 import android.net.Uri
 import android.webkit.CookieManager
+import java.security.cert.X509Certificate
 import kotlin.reflect.KProperty
 
 
@@ -40,7 +43,19 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
         setClientHeaders(options)
         setClientRetryInterceptor(options)
         setClientTimeoutInterceptor(options)
-        addBearerTokenInterceptor(options)
+
+        val bearerTokenInterceptor = getBearerTokenInterceptor(options)
+        if (bearerTokenInterceptor != null) {
+            builder.addInterceptor(bearerTokenInterceptor)
+        }
+
+        val handshakeCertificates = getHandshakeCertificates(options)
+        if (handshakeCertificates != null) {
+            builder.sslSocketFactory(
+                    handshakeCertificates.sslSocketFactory(),
+                    handshakeCertificates.trustManager
+            )
+        }
 
         builder.retryOnConnectionFailure(false);
         builder.addInterceptor(RuntimeInterceptor(this, "retry"))
@@ -58,6 +73,18 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
             for ((k, v) in additionalHeaders.toHashMap()) {
                 clientHeaders.putString(k, v as String)
             }
+        }
+    }
+
+    fun importClientP12(p12FilePath: String, password: String) {
+        val handshakeCertificates = getHandshakeCertificates(p12FilePath, password)
+        if (handshakeCertificates != null) {
+            okHttpClient = okHttpClient.newBuilder()
+                    .sslSocketFactory(
+                        handshakeCertificates.sslSocketFactory(),
+                        handshakeCertificates.trustManager
+                    )
+                    .build()
         }
     }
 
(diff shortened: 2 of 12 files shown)


android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -61,6 +76,18 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
         }
     }
 
+    fun importClientP12(p12FilePath: String, password: String) {
+        val handshakeCertificates = getHandshakeCertificates(p12FilePath, password)
+        if (handshakeCertificates != null) {
+            okHttpClient = okHttpClient.newBuilder()
```
shazm, line 82: Would doing it this way still work for the swizzling? Just wondering if this client will be available via the RCT interceptor.

migbot: Yep, so all we do with the RN client is add an interceptor to it. That interceptor reaches into APIClientModule to find a matching APIClient and if so, uses that APIClient's `okHttpClient` over the RN one. So the APIClient `okHttpClient` can change as much as it wants :)

android/src/main/java/com/mattermost/networkclient/helpers/DocumentHelper.java
```diff
@@ -0,0 +1,212 @@
+package com.mattermost.networkclient.helpers;
```
migbot, line 1: This whole file is basically https://github.com/mattermost/mattermost-mobile/blob/master/android/app/src/main/java/com/mattermost/share/RealPathUtil.java.

enahum: why do you need this?

jupenur (changes requested):
High-level approach looks ok. Spotted one quick thing and will do a more detailed review once dev reviews are done.

android/src/main/java/com/mattermost/networkclient/Extensions.kt
```diff
@@ -46,3 +48,20 @@ fun Request.Builder.applyHeaders(headers: ReadableMap?): Request.Builder {
...
+ */
+fun String.trimTrailingSlashes(): String {
+    return trimEnd { c -> c == '/' }
+}
+
+/**
+ * Computes the MD5 of a string
+ */
+fun String.md5(): String {
+    val md = MessageDigest.getInstance("MD5")
```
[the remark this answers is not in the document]
enahum: @jupenur Would you be able to provide a reason why? I'm very much interested in the answer for this.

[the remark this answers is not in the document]
jupenur: In this case, primarily because every time either we or customers of ours throw the code at some security automation, the automation will complain. There's no good reason to use MD5 other than it being already there in legacy code, so simply not using it makes everyone's lives a little bit easier.

There is a potential concern around hash collisions, too, but how realistic it is depends on how the library is used in the end. Basically, if a new NetworkClient can be created from attacker-controlled input, it's possible an attacker could craft a hash collision which would allow them to mess with the data of other NetworkClients.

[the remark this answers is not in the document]
migbot: OK @jupenur, made the change.

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -61,6 +83,20 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
         }
     }
 
+    fun importClientP12AndRebuildClient(p12FilePath: String, password: String) {
+        importClientP12(p12FilePath, password)
```
enahum, line 87: by reading `buildHandshakeCertificates(options)` it seems that `importClientP12` can throw, don't you need to handle that here?

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -197,14 +235,77 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
...
+        }
+
+        if (heldCertificate != null) {
+            builder.heldCertificate(heldCertificate, *intermediates!!)
+        }
+
+        return builder.build()
+    }
+
+    private fun importClientP12(p12FilePath: String, password: String) {
```
enahum, line 305: Perhaps add that this method throws an exception?

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
```diff
@@ -47,36 +60,117 @@ object KeyStoreHelper {
...
+                    val certificate = p12Store.getCertificate(alias)
+                    val intermediates = p12Store.getCertificateChain(alias)
+
+                    storeP12Contents(key, certificate, intermediates, password, p12Alias)
+
+                    break
+                }
+            }
+        } catch (e: Exception) {
+            Log.d("MIGUEL EXCEPTION", "IMPORTING")
```
enahum, line 87: yeah!!!

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
```diff
@@ -47,36 +60,117 @@ object KeyStoreHelper {
...
+                val certificate = p12Store.getCertificate(certificateAlias) as X509Certificate
+                val heldCertificate = HeldCertificate(KeyPair(certificate.publicKey, key), certificate)
+
+                val intermediates = p12Store.getCertificateChain(keyAlias)
+                        .map { it as X509Certificate }
+                        .toTypedArray()
+
+                return Pair(heldCertificate, intermediates)
+            } else {
+                Log.d("MIGUEL NO CERTIFICATES", keyAlias)
```
enahum, line 114: go go miguel 🎉

enahum (changes requested):
Overall LGTM

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
migbot: I catch where this is called in APIClientModule and reject the promise.

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
migbot: LOL, will remove

android/src/main/java/com/mattermost/networkclient/helpers/KeyStoreHelper.kt
migbot: hahahaha

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
migbot: I did a quick search and didn't see anything like Swift's `throws` but will do a better search today.

migbot: No dice. `importClientP12` is called in two places: 1) when a p12 is passed during client creation via `sessionConfiguration` and 2) when importing on the JS side via `client.importClientP12For`. In the first case we don't want the client creation to proceed even if there's an error when processing the p12, so in this case we just [send a client error event back to JS](https://github.com/mattermost/react-native-network-client/pull/71/files#diff-3f3de389fcdc8d34a55e7f21d4a0a5eae7c72f8f4b1cff376543c38e822efa52R275), while in the second case we do want to reject the promise so we allow the exception to bubble up to [APIClientModule.importClientP12For](https://github.com/mattermost/react-native-network-client/pull/71/files#diff-607ad46bbe32e3276def3975236b17bc83c01570c8af248bc3573730ab5fff2aR166).

migbot: Added a comment to the function, though.

enahum (approved):

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -11,7 +11,8 @@
       android:allowBackup="false"
       android:theme="@style/AppTheme"
       android:networkSecurityConfig="@xml/network_security_config"
-      android:usesCleartextTraffic="true">
+      android:usesCleartextTraffic="true"
+      android:requestLegacyExternalStorage="true">
```
enahum, line 15: Wondering if we should be using the new permission system for this.. for now it's fine but..

enahum (approved):

example/android/app/src/main/AndroidManifest.xml
migbot: Yep looked through https://github.com/react-native-image-picker/react-native-image-picker/issues/1233, https://github.com/zoontek/react-native-permissions/pull/587, and a few others and it looks like this is the way to go for now.

josephbaylon (approved):
@migbot This generally works for `put/patch/post/delete/get/fast image` but I'm not able to test secure file uploads and secure websocket because of existing issues for those two (see below). I can retest secure connection on a different PR once those two are fixed. Approving this PR.

https://mattermost.atlassian.net/browse/MM-35347
https://mattermost.atlassian.net/browse/MM-35349

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -197,14 +235,79 @@ class NetworkClient(private val baseUrl: HttpUrl? = null, private val options: R
...
+        // KeyStoreHelper.importClientCertificateFromP12 can throw an exception
+        // and we leave it up to the caller of `importClientP12` to handle it.
```
jupenur, line 307: Is there a reason for not using KDoc here?

jupenur (approved):
No security concerns beyond what was already communicated.

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
migbot: Just my ignorance!

PR #74, opened 2021-05-06
MM-35348 Android unit tests for ExponentialRetryInterceptor

migbot:
#### Summary
* Fixes sleep interval by using passing milliseconds
* Tests that ExponentialRetryInterceptor calls `getWaitInterval` and `waitForMilliseconds` with the expected arguments and for the expected times

https://mattermost.atlassian.net/browse/MM-35348


android/build.gradle
```diff
@@ -129,4 +129,8 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "com.squareup.okhttp3:okhttp:4.9.1"
   implementation "com.squareup.okhttp3:okhttp-tls:4.9.1"
+
+  testImplementation "junit:junit:4.12"
```
enahum, line 133: 😱

example/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed May 05 14:13:34 MST 2021
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-bin.zip
```
enahum, line 6: Why? And how it affects other RN apps?

enahum (commented):
Definitely not against it, just want to know why the native tests instead of the e2e tests?

example/android/gradle/wrapper/gradle-wrapper.properties
migbot: > Definitely not against it, just want to know why the native tests instead of the e2e tests?

To follow AF in that the retry implementation is tested on the native side. The JS side need only test that `retriesExhausted` is included when expected.

android/src/main/java/com/mattermost/networkclient/interceptors/ExponentialRetryInterceptor.kt
```diff
@@ -17,6 +17,6 @@ class ExponentialRetryInterceptor(
     }
 
     override fun getWaitInterval(attempts: Int): Long {
-        return (exponentialBackoffBase.pow(attempts) * exponentialBackoffScale).toLong()
+        return (exponentialBackoffBase.pow(attempts) * exponentialBackoffScale * 1000).toLong()
```
josephbaylon, line 20: @migbot Should we do the same on iOS?

example/android/gradle/wrapper/gradle-wrapper.properties
migbot: Ah and the gradle update was needed when I was trying to mock context. Let me remove this.

migbot, line 6:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
```

android/src/main/java/com/mattermost/networkclient/interceptors/ExponentialRetryInterceptor.kt
josephbaylon: Or to make it consistent with linear retry, exponentialBackoffBase and exponentialBackoffScale fields in example app to be settable in millis?

migbot: Not needed on iOS as the `retryWithDelay` function expects seconds. For consistency, I can use `TimeUnit.SECONDS.sleep` instead of `TimeUnit.MILLISECONDS.sleep` and not multiply by 1000 here. I'll make the change now.

migbot: Actually, nevermind. We'll keep `TimeUnit.MILLISECONDS.sleep` since the LinearRetryInterceptor passes in milliseconds. I just refactored this line to be a bit more clear.

enahum (approved):

josephbaylon (changes requested):
@migbot Looks like `put/patch/post/delete` work fine on exponential retry but `get` returns error overlay instead of response overlay

Screen Shot 2021-05-06 at 1 19 08 PM

example/android/build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:3.5.3"
+        classpath "com.android.tools.build:gradle:4.1.0"
```
migbot, line 16:
```suggestion
        classpath "com.android.tools.build:gradle:3.5.3"
```

josephbaylon (approved):
@migbot I'll approve it since it's working on your end. I'll keep digging and just submit a server fix if I find the issue. Odd though that it's only failing on `get` for me.

PR #78, opened 2021-05-12
MM-35350 Disable error response overlay tests on android

josephbaylon:
#### Summary
- Since errors won't be consistent on both platforms, error overlay tests will just be run against iOS and not on Android

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35350


enahum (approved):

josephbaylon:
@migbot Enabled check of fields on anroid

migbot (approved):

PR #100, opened 2021-10-19
fix: force unwrap on UploadRequestBody.kt and ApiClient.swift

avinashlng1080:
#### Summary
We are using this library on the Mattermost mobile app and we noticed a crash happening on both iOS and Android when performing image uploads.

The cause of the crash was due to some optional values being force-unwrapped and thus causing a Null Pointer exception.

This PR fixes this issue on both platforms.


enahum:
Nice description, you do know we own this library, right?

enahum (approved):

avinashlng1080:
Yes, I do :)

PR #105, opened 2022-07-21
fix crash on Android when default locale format numbers with comma

vovkasm:
#### Summary
This fixed crash on Android with non english numeric locale.

See this SO answer and corresponding question: https://stackoverflow.com/a/49432344

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/helpers/ProgressListener.kt
+++ b/android/src/main/java/com/mattermost/networkclient/helpers/ProgressListener.kt
@@ -2,8 +2,7 @@ package com.mattermost.networkclient.helpers
 
 import com.facebook.react.bridge.Arguments
 import com.mattermost.networkclient.APIClientModule
-import java.math.RoundingMode
-import java.text.DecimalFormat
+import kotlin.math.roundToInt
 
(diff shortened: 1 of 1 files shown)


enahum:
Hey @vovkasm thanks for this a couple of questions:
1. Would you mind signing the CLA so we can eventually merge this PR
2. What do you think about this suggestion instead of your approach
```kotlin
// 2-Decimal places, rounded up
private val df = DecimalFormat("#.00")
    .apply{
        roundingMode = RoundingMode.UP
        decimalFormatSymbols = DecimalFormatSymbols(Locale.US)
    }
```

vovkasm:
@enahum 

> 1. Would you mind signing the CLA so we can eventually merge this PR

Yes, I would like to!

> What do you think about this suggestion instead of your approach

I think that using complex formatting and parsing algorithms for just for rounding numbers are not the most efficient use of mobile processor resources ;-)

enahum (approved):

enahum:
@avinashlng1080 can you make sure this PR is merged and the reference in the mobile app is updated once @vovkasm signs the CLA?

Thanks.

PR #108, opened 2022-12-13
feat: clear request cache on client invalidate

enahum:
#### Summary
On client invalidate remove the entries from the cache.

on Android: if there are any cache entries that match the domain, we remove them all when the client is invalidated.
on iOS: There is no easy way to remove only entries that match a specific domain, we can only retrieve entries for a specific URLRequest or Task, so on client invalidate we are removing all cache responses.


larkox (approved):

enahum:
not needed

PR #109, opened 2022-12-16
Fix not properly closing the websocket when starting airplane mode

larkox:
#### Summary
Some of the changes here are speculative, but after testing they seem to work just fine.

On the android side, the main problem is that we were receiving a `SocketException` instead of the other exceptions we were checking for. After a quick look, it seems safe to consider any error here probably is enough to close the connection. https://docs.oracle.com/javase/7/docs/api/java/net/SocketException.html

On the iphone side, the error is a bit more criptic, since it says not much more than error 57. After a big of googling, it seems to be the "websocket not connected" error, which also seems enough to close the websocket.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48802


enahum (approved):

enahum:
I'll release a new version of the library in an hour or so

PR #114, opened 2023-03-03
Reject promise if we try to create a ws client when we already have one

larkox:
#### Summary
If we try to create a second websocket client too fast, there may be a race condition on the js side. By rejecting here, we avoid these issues.

#### Ticket Link
Related to: https://github.com/mattermost/mattermost-mobile/pull/7185

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
+++ b/android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
@@ -53,6 +53,10 @@ internal class WebSocketClientModule(reactContext: ReactApplicationContext) : Re
             return promise.reject(error)
         }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/mattermost/networkclient/WebSocketClientModule.kt
```diff
@@ -53,6 +53,10 @@ internal class WebSocketClientModule(reactContext: ReactApplicationContext) : Re
             return promise.reject(error)
         }
 
+        if (clients.containsKey(wsUri)) {
```
enahum, line 56: What about ios?

enahum (approved):

hmhealey (approved):

PR #120, opened 2023-07-11
Listen for timeout options on upload

larkox:
#### Summary
If we pass the timeout option to and upload request, on Android was not listening, stopping it at 30 seconds.

This makes sure we are listening to this.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-53445
Fix https://github.com/mattermost/mattermost-mobile/issues/5818


enahum (approved):

enahum:
does `download` has the same problem? or is just `upload` ?

larkox:
just `upload`. `download` seems to properly set this timeout a few lines below. Also, it only affects Android. iOS seems to properly be looking into the options to set the timeout for every request.

PR #121, opened 2023-07-11
chore: tweak library defaults to be more consistent with what React Native does

larkox:
#### Summary
Tweak defaults to be more consistent with what React Native does. This may solve some problems with timeouts and request parallelization.

#### Ticket Link
May solve https://mattermost.atlassian.net/browse/MM-49968
May solve https://github.com/mattermost/mattermost-mobile/issues/7021

### DIFF
--- a/example/src/hooks/index.tsx
+++ b/example/src/hooks/index.tsx
@@ -28,8 +28,8 @@ export const useSessionConfiguration = (): UseSessionConfigurationResponse => {
     ] = useState<SessionConfiguration>({
         allowsCellularAccess: true,
         waitsForConnectivity: false,
-        timeoutIntervalForRequest: 30000,
-        timeoutIntervalForResource: 30000,
+        timeoutIntervalForRequest: 60000,
+        timeoutIntervalForResource: 604800000, // 7 days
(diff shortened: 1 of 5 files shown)


example/src/hooks/index.tsx
```diff
@@ -28,8 +28,8 @@ export const useSessionConfiguration = (): UseSessionConfigurationResponse => {
     ] = useState<SessionConfiguration>({
         allowsCellularAccess: true,
         waitsForConnectivity: false,
-        timeoutIntervalForRequest: 30000,
-        timeoutIntervalForResource: 30000,
+        timeoutIntervalForRequest: 60000,
+        timeoutIntervalForResource: 604800000, // 7 days
```
enahum, line 32: should we keep it undefined so that the OS applies the default values instead?

enahum (approved):

example/src/hooks/index.tsx
larkox: This is also used on android apparently (`android/src/main/java/com/mattermost/networkclient/NetworkClientBase.kt`). So... that part will also get the default one.

So... I guess yes, that is the better approach.

PR #122, opened 2023-08-09
fix: handle exception for Android timeouts config

enahum:
#### Summary
on Android if the timeouts for resource and request was passed as undefined, the creation of the client fails. in that case we will set it to 0 so that is set to no timeout


larkox (approved):

larkox:
Now that I think of it... instead of no timeouts, shouldn't it be a default value? (not sure which one)

PR #123, opened 2023-09-05
feat: emit event when server certificate is not trusted

enahum:
#### Summary
When the server certificate (SSL) is not trusted or invalid we now emit an event so that the app can show an error message

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54285

### DIFF
--- a/ios/APIClient.swift
+++ b/ios/APIClient.swift
@@ -13,12 +13,14 @@ import React
 
 enum APIClientError: Error {
     case ClientCertificateMissing
+    case ServerCertificateInvalid
 }
 
 extension APIClientError: LocalizedError {
     var errorCode: Int {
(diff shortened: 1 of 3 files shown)


ios/APIClient.swift
```diff
@@ -13,19 +13,23 @@ import React
...
+        case .ServerCertificateInvalid: return -299
         }
     }
     
     var errorDescription: String? {
         switch self {
         case .ClientCertificateMissing:
             return "Failed to authenticate: missing client certificate"
+        case .ServerCertificateInvalid:
+            return "Invalid or not trusted server certificate"
```
enahum, line 32: this string is irrelevant

larkox (approved):

PR #126, opened 2024-04-04
chore: dependencies update

larkox:
#### Summary
Update dependencies for the library and example app

#### Ticket Link
NONE


example/ios/patches/SessionDelegate.patch
enahum, line 1: This looks incorrect

enahum (approved):
My bad.. patch looks good

PR #128, opened 2024-06-10
support New arch

enahum:
#### Summary
Add support for new architecture with backwards compatibility


example/ios/.xcode.env.local
```diff
@@ -0,0 +1,2 @@
+export NODE_BINARY=/Users/enahum/.nvm/versions/node/v18.17.1/bin/node
```
larkox, line 1: Unintended commit?

ios/NetworkClient-Bridging-Header.h
```diff
@@ -1,18 +1 @@
-//
-//  NetworkClient-Bridging-Header.h
-//  NetworkClient
-//
-//  Created by Miguel Alatzar on 12/2/20.
-//  Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
-//  See LICENSE.txt for license information.
-//
```
larkox, line 1: This ends up being an empty file. Is that intended?

larkox (approved):
Great work! Just a couple of minor comments.

PR #134, opened 2024-11-01
perf(android): add a cache object to avoid having to decrypt the alias in every access

enahum:
#### Summary
On Android, we were decrypting the bearer token for every request we were making, and this was causing problems specially on Android 15. With this PR, we created a map that will hold the token in memory for quicker access without having to decrypt every time.


larkox (approved):

rahimrahman (approved):
If we're doing hundreds of requests (especially when app comes back to foreground), this could amount to a significant improvement.

PR #135, opened 2024-12-03
feat: collect metrics for requests

enahum:
#### Summary
This PR introduces the capability to collect metrics for requests in a session

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61877

### DIFF
--- a/ios/NetworkClient.swift
+++ b/ios/NetworkClient.swift
@@ -1,5 +1,7 @@
 import Alamofire
 import SwiftyJSON
+import Network
+import CoreTelephony
 
 let RETRY_TYPES = ["EXPONENTIAL_RETRY": "exponential", "LINEAR_RETRY": "linear"]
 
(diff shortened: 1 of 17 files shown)


ios/NetworkClient.swift
```diff
@@ -293,4 +364,62 @@ extension NetworkClient {
...
+        monitor.pathUpdateHandler = { path in
+            if path.usesInterfaceType(.wifi) {
+                networkType = "Wi-Fi"
+            } else if path.usesInterfaceType(.cellular) {
+                let telephonyNetworkInfo = CTTelephonyNetworkInfo()
+                if let currentRadioAccessTechnology = telephonyNetworkInfo.serviceCurrentRadioAccessTechnology?.values.first {
+                    if #available(iOS 14.1, *) {
+                        switch currentRadioAccessTechnology {
+                        case CTRadioAccessTechnologyGPRS:
+                            networkType = "2G (GPRS)"
```
isacikgoz, line 383: Is it possible to run this application on 2G networks?

isacikgoz (approved):
LGTM overall. wondering if we could classify network regarding to their speed rather than their type?

ios/NetworkClient.swift
enahum: I don't think it will happen but I see no reason why it shouldn't work

enahum:
> LGTM overall. wondering if we could classify network regarding to their speed rather than their type?

I'll give it a try and see if I can add it, but no promises

src/APIClient/NativeApiClient.ts
```diff
@@ -6,6 +6,18 @@ import type {
...
+    networkType: string;
+    tlsCipherSuite: string;
+    tlsVersion: string;
+    httpVersion: string;
+    isCached: boolean;
+    compressedSize: Double;
+    size: Double;
+    connectionTime: Double;
+    latency: Double;
+};
```
rahimrahman, line 19: Shouldn't this be the same as APIClient.ts?  Can't we define here and re-export there?

Missing at least `speedInMbps`, maybe?

rahimrahman (approved):
I must admit, I asked chatgpt to help review this and couldn't find anything worth bringing up.  I used your branch for testing with multiple scenarios (increasing latency).

I only have 1 comment about missing fields in a type in which I think could be the same for both for codegen and APIclient.

src/APIClient/NativeApiClient.ts
enahum: fixed

PR #138, opened 2024-12-16
Migrate SharedPreferences to DataStore

enahum:
#### Summary
On Android we were using SharedPreferences to store the client certs and the session tokens but now the recommended way is to use [DataStore](https://developer.android.com/topic/libraries/architecture/datastore) instead

This PR migrates the current SharedPreferences to DataStore and starts using DataStore from this point onwards.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -64,4 +64,5 @@ dependencies {
   implementation "com.squareup.okhttp3:okhttp:4.12.0"
   implementation "com.squareup.okhttp3:okhttp-tls:4.12.0"
   implementation "com.squareup.okhttp3:okhttp-urlconnection:4.12.0"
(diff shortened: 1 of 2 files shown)


enahum:
@larkox @rahimrahman I would like to have this reviewed soon so that I can merge and release a new version of this library.

Thanks boys.

larkox (approved):

enahum:
> Looks good, wondering about runBlocking (why would there be a need to block), and if there should be any try..catch?

@rahimrahman datastore is async, perhaps check the document I linked?

rahimrahman:
> > Looks good, wondering about runBlocking (why would there be a need to block), and if there should be any try..catch?
> 
> @rahimrahman datastore is async, perhaps check the document I linked?

Thanks.  Rephrase.  Document you linked says to avoid.

PR #149, opened 2025-09-12
Remove incorrect certificate warning

Jonty:
#### Summary
`javax.net.ssl.SSLHandshakeException` can be thrown for a variety of reasons unrelated to the certificate being invalid - most notably socket errors while trying to negotiate the TLS session. This can result in spurious warnings about the remote host serving an invalid certificate when a connection is closed remotely, and can easily be triggered by a network change.

Given the certificate is validated elsewhere we can trust that users will be warned appropriately so removing this unnecessary warning path is the cleanest fix - handling all relevant sub-errors would be very ugly.

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
+++ b/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
@@ -245,9 +245,6 @@ internal class NetworkClient(private val context: Context, private val baseUrl: 
                         rejectInvalidCertificate(promise, request.url.host)
                         return
                     }
-                } else if (e is javax.net.ssl.SSLHandshakeException) {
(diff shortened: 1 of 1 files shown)


Jonty:
Would it be possible for someone to take a look at this? The issue still occurs sporadically and I've heard nothing in response since signing the CLA.

jchia:
This also affects diagnostics in Mattermost Mobile. The current SSLHandshakeException branch calls rejectInvalidCertificate, replacing the original exception with a generic java.lang.Exception/EUNSPECIFIED. Consequently, callers such as getServerUrlAfterRedirect cannot log the underlying TLS failure. Allowing the exception to reach promise.reject(e), as this PR does, preserves the actionable error and avoids the misleading certificate classification.

PR #158, opened 2026-03-20
ci: add TypeScript and unit test jobs, fix pre-existing CI failures

pavelzeman:
## Summary

Builds on #161 (now merged). Adds the remaining two CI jobs — **TypeScript** and **unit tests** — along with fixes for pre-existing issues that prevented them from passing.

### Jobs added
- **TypeScript** — `npm run typescript` (tsc --noEmit)
- **Unit Tests** — `npm test -- --coverage --passWithNoTests`

### Fixes

**TypeScript errors (40+ failures):** All from `example/src/` referencing React Native dependencies (`@react-navigation/stack`, `react-native-elements`, etc.) that are only installed in `example/node_modules`, not at the root. Fix: simplify `tsconfig.json` exclude to just `example` instead of individual subdirectories.

**Unit Tests "No tests found" (exit code 1):** This repo has no JS/TS test files — all tests are Kotlin tests in `test-runner/`. Jest exits with code 1 when it finds zero matching test files. Fix: add `--passWithNoTests` to the Jest invocation.

### Build dependency
`build` job now gates on all four check jobs: lint, typescript, unit-tests, kotlin-tests.

### Notes
- Node.js 22 (aligned with #161)
- Actions pinned to full commit SHAs
- CircleCI config left in place for now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,30 @@ jobs:
       - run: npm ci
       - run: npm run lint
 
+  typescript:
+    name: TypeScript
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: npm run typescript
+
+  unit-tests:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: npm test -- --coverage --passWithNoTests
+
   kotlin-tests:
     name: Kotlin Tests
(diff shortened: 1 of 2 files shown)


esarafianou (commented):
Since there are zero JS/TS test files in this repo, Jest will always exit immediately having done nothing. This burns ~60-90s of CI time on every PR and push to master for a no-op.

I'd recommend adding this job when we introduce JS/TS test files so CI time isn't wasted until then. @pavelzeman did you intentionally include it now so the pipeline is ready when tests are added? That's a valid approach too, it's mainly the CI cost that steers me towards skipping it for now.

larkox (approved):

pavelzeman:
Good call @esarafianou — you're right, there are zero JS/TS test files in this repo so the Jest job was a pure no-op burning CI time. Removed it in c4d4515. We can add it back when JS/TS tests are actually introduced.

Also cleaned up the redundant `tsconfig.json` exclude entries per CodeRabbit's nitpick.

pavelzeman:
### CI Jobs Summary

For clarity on what each job actually does:

- **Lint** — ESLint static analysis
- **TypeScript** — `tsc --noEmit` type-checking (not tests — validates the source compiles cleanly; this caught 40+ real type errors from `example/` leaking into the check)
- **Kotlin Tests** — the only actual test suite in this repo (5 tests: OkHttp interceptor behavior via MockWebServer)
- **Build** — `npm run prepare` (compiles the publishable package)

No JS/TS test files exist in this repo, so there's no Jest job. Can be added when JS/TS tests are introduced.

esarafianou (approved):

PR #159, opened 2026-03-20
Add Kotlin unit test infrastructure and CircleCI step

pavelzeman:
#### Summary

Add infrastructure for running Kotlin unit tests in CI. Currently the repo has one Kotlin test file (`ExponentialRetryInterceptorTest.kt`) but no test dependencies in `build.gradle` and no CI step to run them.

The Android side of this library has **36 Kotlin files totaling ~3,250 lines of code** — including interceptors for auth, retry, compression, timeouts, and metrics. That's a decent amount of logic with no test coverage in CI. This PR fixes that.

#### Ticket Link

N/A — infrastructure improvement to support ongoing Sentry issue fixes.

#### Release Note

```release-note
NONE
```

---

### Changes

- **`android/build.gradle`**: Add `testImplementation` dependencies for JUnit 4, Mockito, and OkHttp MockWebServer
- **`test-runner/`**: Standalone Gradle project with wrapper for running Kotlin unit tests independently of the React Native build. Includes a smoke test to verify the infrastructure works.
- **`.circleci/config.yml`**: Add `kotlin-tests` job using `cimg/android` executor, wired as a gate for `build-package`

### Why a standalone test-runner?

The Android module (`android/build.gradle`) is a library that depends on React Native at build time, making it difficult to run `./gradlew test` standalone. The `test-runner` project provides a lightweight Kotlin+JVM environment with OkHttp dependencies — enough to test interceptor logic without the full React Native build chain.

### What this enables

- TDD for Android interceptor code (e.g. header validation, retry logic)
- CI enforcement of Kotlin test correctness
- Foundation for upcoming interceptor bug fixes ([#155](https://github.com/mattermost/react-native-network-client/pull/155))

---
_Migrated from #156 (fork → org branch)_

PR #160, opened 2026-03-20
Fix OkHttp header validation crashes from non-ASCII characters on Android

pavelzeman:
## Summary

Fixes two OkHttp `IllegalArgumentException` crashes on Android where header values contain characters that fail OkHttp's ASCII validation.

### 1. CompressedResponseSizeInterceptor — Arabic locale digits (MATTERMOST-MOBILE-ANDROID-AZ02)

On devices with Arabic system locale, `String.format("%.4f", value)` produces Arabic-Indic digits (U+0660–U+0669) in the `X-Speed-Mbps` header. OkHttp rejects these as non-ASCII.

**Fix:** Use `String.format(Locale.US, "%.4f", value)` to ensure ASCII digits regardless of device locale.

### 2. applyHeaders — corrupted token with control characters (MATTERMOST-MOBILE-ANDROID-AYPW)

When the JS layer passes a token containing non-ASCII control characters (e.g. `0x02`), OkHttp throws `IllegalArgumentException` during `addHeader()` in `applyHeaders()` at request construction time — before any interceptor runs.

**Fix:** Catch `IllegalArgumentException` per-header in `applyHeaders()`, log a warning, skip the invalid header, and let the request proceed. The server will respond with a 4xx that the JS layer can handle gracefully. As a transport layer, this library should not silently modify or sanitize credentials.

A follow-up investigation into why the JS layer sends corrupted tokens is warranted.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/Extensions.kt
+++ b/android/src/main/java/com/mattermost/networkclient/Extensions.kt
@@ -1,5 +1,6 @@
 package com.mattermost.networkclient
 
+import android.util.Log
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableArray
@@ -134,8 +135,17 @@ fun Response.toDownloadMap(path: String): WritableMap {
 fun Request.Builder.applyHeaders(headers: Map<String, Any?>?): Request.Builder {
     if (headers != null){
         for ((k, v) in headers) {
-            this.removeHeader(k)
-            this.addHeader(k, v.toString())
+            try {
+                this.removeHeader(k)
+                this.addHeader(k, v.toString())
+            } catch (e: IllegalArgumentException) {
+                // OkHttp validates header values and rejects non-ASCII characters
+                // (e.g. control chars like 0x02 in corrupted auth tokens, or
+                // Arabic-Indic digits from locale-dependent formatting).
+                // Skip the invalid header and let the request proceed — the server
+                // will reject it with a 4xx that the JS layer can handle gracefully.
+                Log.w("NetworkClient", "Skipping header '$k': ${e.message}")
+            }
         }
     }
 
--- a/android/src/main/java/com/mattermost/networkclient/interceptors/CompressedResponseSizeInterceptor.kt
+++ b/android/src/main/java/com/mattermost/networkclient/interceptors/CompressedResponseSizeInterceptor.kt
@@ -3,6 +3,7 @@ package com.mattermost.networkclient.interceptors
 import okhttp3.Interceptor
 import okhttp3.Response
 import okhttp3.ResponseBody.Companion.toResponseBody
+import java.util.Locale
 
 class CompressedResponseSizeInterceptor: Interceptor {
     override fun intercept(chain: Interceptor.Chain): Response {
@@ -37,7 +38,9 @@ class CompressedResponseSizeInterceptor: Interceptor {
             .header("X-Compressed-Size", compressedSize.toString())
             .header("X-Start-Time", startTime.toString())
             .header("X-End-Time", endTime.toString())
-            .header("X-Speed-Mbps", "%.4f".format(speedMbps)) // Format to 4 decimal places
+            // Use Locale.US to ensure ASCII digits in header values.
+            // System locale formatting (e.g. Arabic) produces non-ASCII digits that OkHttp rejects.
+            .header("X-Speed-Mbps", String.format(Locale.US, "%.4f", speedMbps))
             .build()
     }
 }
--- a/test-runner/src/test/kotlin/com/mattermost/networkclient/ApplyHeadersTest.kt
+++ b/test-runner/src/test/kotlin/com/mattermost/networkclient/ApplyHeadersTest.kt
@@ -1,0 +1,177 @@
+package com.mattermost.networkclient
+
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.Interceptor
+import okhttp3.Response
+import org.junit.Assert
+import org.junit.Test
+
+/**
+ * Tests for applyHeaders resilience against invalid header values.
+ *
+ * Reproduces MATTERMOST-MOBILE-ANDROID-AYPW: corrupted tokens containing
+ * non-ASCII control characters (e.g. 0x02) cause OkHttp to throw
+ * IllegalArgumentException in addHeader() during request construction.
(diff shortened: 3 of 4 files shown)


android/src/main/java/com/mattermost/networkclient/interceptors/BearerTokenInterceptor.kt
```diff
@@ -12,9 +12,15 @@ class BearerTokenInterceptor(private val alias: String, private val bearerAuthTo
 
         var token = ApiClientModuleImpl.retrieveValue(alias)
         if (token !== null) {
-            request = request.newBuilder()
-                    .header("Authorization", "Bearer $token")
-                    .build()
+            // Sanitize token to ASCII printable characters only.
+            // Corrupted tokens with control characters (e.g. 0x02) cause OkHttp to throw
+            // IllegalArgumentException when setting the Authorization header.
+            val sanitizedToken = token.replace(Regex("[^\\x20-\\x7E]"), "")
```
Willyfrog: nitpick: If sanitizedToken and token are different, we should log it.

if there is ever an issue, we might go crazy if the token is being modified and we don't know about it.

Willyfrog (approved):
LGTM, added a nitpick to prevent headaches when debugging issues in the future

android/src/main/java/com/mattermost/networkclient/interceptors/BearerTokenInterceptor.kt
pavelzeman: Good call — added a `Log.w` when the token is sanitized. Logs the number of removed chars without exposing the token value itself. 9eb212d

test-runner/src/test/kotlin/com/mattermost/networkclient/BearerTokenSanitizationTest.kt
```diff
@@ -0,0 +1,155 @@
...
+            val sanitizedToken = token.replace(Regex("[^\\x20-\\x7E]"), "")
+            if (sanitizedToken.isEmpty()) {
+                return chain.proceed(chain.request())
+            }
+            val request = chain.request().newBuilder()
+                .header("Authorization", "Bearer $sanitizedToken")
+                .build()
+            return chain.proceed(request)
+        }
+    }
```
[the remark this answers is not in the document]
pavelzeman: By design — the standalone test-runner can't compile the production interceptor because it depends on `ApiClientModuleImpl` which imports React Native (`com.facebook.react.bridge`). The test-runner is a pure JVM/Kotlin environment without Android/RN dependencies. The duplicated logic is intentional to validate the sanitization pattern in isolation.

larkox (approved):
I would love some security eyes on this PR. Probably I am too cautious, but if someone from @esarafianou team can take a look to see whether what we are doing is correct, I would feel more comfortable.

Also, similar to the other PR. I love we adding tests to the native side, but not sure if we are running these tests on CI. A great improvement if we start doing it.

pavelzeman:
Leaving open a little longer, so that @esarafianou or @edgarbellot can have a look.
The Kotlin CI this depends on was already merged - https://github.com/mattermost/react-native-network-client/pull/159

esarafianou (changes requested):
The changes in `CompressedResponseSizeInterceptor.kt` for [MATTERMOST-MOBILE-ANDROID-AZ02](https://mattermost-mr.sentry.io/issues/7194769442/) look good. 

I don't think that the `BearerTokenInterceptor.kt` changes address the [MATTERMOST-MOBILE-ANDROID-AYPW](https://mattermost-mr.sentry.io/issues/7077385609/) crash. Looking at the Sentry stack trace, the crash happens in `ExtensionsKt.applyHeaders` at line 138, called from `NetworkClient.buildRequest` at line 446.  That's the per-request headers path during request construction.

Request construction and interceptor execution are two separate phases. When `buildRequest` runs, it calls `applyHeaders`, which calls OkHttp's `addHeader()` and OkHttp validates the header value immediately. If the value contains a non-ASCII character like 0x02, it throws `IllegalArgumentException` and the request never finishes being built. `BearerTokenInterceptor` is an OkHttp interceptor, which means it only runs after the request is fully constructed and handed to the interceptor chain for execution. It never gets a chance to sanitize anything because the crash has already happened during construction.

The fix should go in `applyHeaders` in `Extensions.kt`, but I don't think it should sanitize the header value. This library is a transport layer, it shouldn't silently modify credentials or decide what a good enough token looks like. Instead, `applyHeaders` should catch the `IllegalArgumentException` from `addHeader`, skip the invalid header, and let the request proceed without it. This turns an unhandled crash into a failed request that the JS layer can handle gracefully. 

I'd be also worth opening a follow up to investigate why the JS layer is sending a token with 0x02 in upload request headers. This library is just the transport layer, so the real fix should be upstream wherever the token gets corrupted.

cc. @larkox, @Willyfrog tagging you for visibility since you're our mobile experts

pavelzeman (commented):
Thanks @esarafianou — great catch on the interceptor vs request construction timing. You're absolutely right: the crash happens in `applyHeaders` during `addHeader()`, well before any interceptor in the chain gets a chance to run. The `BearerTokenInterceptor` fix was addressing the wrong phase entirely.

Reworked the PR based on your feedback:

- **Reverted** `BearerTokenInterceptor.kt` to its original state
- **Moved the fix** to `applyHeaders()` in `Extensions.kt` — catches `IllegalArgumentException` per-header, logs a warning, skips the invalid header, and lets the request proceed
- **No sanitization** — as you noted, this is a transport layer and shouldn't silently modify credentials. The server will reject with a 4xx that JS can handle
- **Updated tests** to validate the new approach (catch-and-skip in applyHeaders instead of token sanitization)

We're also investigating upstream why the JS layer is sending a token with `0x02` — will open a follow-up issue once we have findings.

pavelzeman (commented):
Thanks @esarafianou — great catch on the interceptor vs request construction timing. You're absolutely right: the crash happens in `applyHeaders` during `addHeader()`, well before any interceptor in the chain gets a chance to run. The `BearerTokenInterceptor` fix was addressing the wrong phase entirely.

Reworked the PR based on your feedback:

- **Reverted** `BearerTokenInterceptor.kt` to its original state
- **Moved the fix** to `applyHeaders()` in `Extensions.kt` — catches `IllegalArgumentException` per-header, logs a warning, skips the invalid header, and lets the request proceed
- **No sanitization** — as you noted, this is a transport layer and shouldn't silently modify credentials. The server will reject with a 4xx that JS can handle
- **Updated tests** to validate the new approach (catch-and-skip in applyHeaders instead of token sanitization)

We're also investigating upstream why the JS layer is sending a token with `0x02` — will open a follow-up issue once we have findings.

esarafianou (approved):

enahum (approved):

PR #163, opened 2026-03-26
[MM-67999] fix websocket crash

Willyfrog:
#### Summary

this is the second part of the websocket crash. See: https://github.com/mattermost/mattermost-mobile/pull/9639

fix(ios): synchronize WebSocket dictionary access and prevent delegate use-after-free

Add a serial dispatch queue to protect the `webSockets` dictionary in WebSocketManager from concurrent access across URLSession delegate callbacks, main queue, and the RN bridge. All reads/writes are now wrapped in `queue.sync {}` with atomic remove-and-return in `invalidateClient` and atomic snapshot+removeAll in `invalidateContext` to eliminate TOCTOU races.

In WebSocketWrapper.didReceive, capture the weak `self.delegate` into a strong local at method entry so the delegate cannot deallocate mid-callback, preventing EXC_BAD_ACCESS.

Addresses Sentry issues MATTERMOST-MOBILE-IOS-XHW and MATTERMOST-MOBILE-IOS-YQG.

Once this is done, we'll need to update the dependencies on the app
#### Ticket Link

[MM-67999](https://mattermost.atlassian.net/browse/MM-67999)

### DIFF
--- a/ios/WebSocket/WebSocketManager.swift
+++ b/ios/WebSocket/WebSocketManager.swift
@@ -5,19 +5,14 @@ import SwiftyJSON
 class WebSocketManager: NSObject {
     static let `default` = WebSocketManager()
     private override init() {}
-    internal var webSockets: [URL: WebSocket] = [:]
-    
+    private var webSockets: [URL: WebSocket] = [:]
+    private let queue = DispatchQueue(label: "com.mattermost.react-native-network-client.websocket-manager")
+
     func webSocketCount() -> Int {
-        return webSockets.count
+        return queue.sync { webSockets.count }
     }
-    
+
     func createWebSocket(for url:URL, withOptions options: Dictionary<String, Any>, withDelegate delegate: WebSocketWrapper) throws -> Void {
-        let existingWebSocket = getWebSocket(for: url)
-        if (existingWebSocket != nil) {
-            existingWebSocket!.delegate = delegate
-            return
-        }
-        
         var request = URLRequest(url: url)
         var compressionHandler: CompressionHandler? = nil
         var clientCredential: URLCredential? = nil
@@ -30,15 +25,15 @@ class WebSocketManager: NSObject {
                     request.addValue(value.stringValue, forHTTPHeaderField: key)
                 }
             }
-            
+
             if let timeoutInterval = options["timeoutInterval"].double {
                 request.timeoutInterval = timeoutInterval / 1000
             }
-            
+
             if let clientP12Configuration = options["clientP12Configuration"].dictionaryObject as? [String:String] {
                 let path = clientP12Configuration["path"]
                 let password = clientP12Configuration["password"]
-                
+
                 do {
                     try Keychain.importClientP12(withPath: path!, withPassword: password, forHost: url.host!)
                 } catch KeychainError.DuplicateIdentity {
@@ -50,7 +45,7 @@ class WebSocketManager: NSObject {
                 let (identity, certificate) = try Keychain.getClientIdentityAndCertificate(for: url.host!)!
                 clientCredential = URLCredential(identity: identity, certificates: [certificate], persistence: URLCredential.Persistence.permanent)
             }
-            
+
             if options["trustSelfSignedServerCertificate"].boolValue {
                 allowSelfSign = true
             }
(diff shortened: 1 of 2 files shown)


ios/WebSocket/WebSocketManager.swift
```diff
@@ -5,19 +5,14 @@ import SwiftyJSON
...
+        return queue.sync { webSockets.count }
     }
-    
+
     func createWebSocket(for url:URL, withOptions options: Dictionary<String, Any>, withDelegate delegate: WebSocketWrapper) throws -> Void {
-        let existingWebSocket = getWebSocket(for: url)
-        if (existingWebSocket != nil) {
-            existingWebSocket!.delegate = delegate
-            return
-        }
```
enahum, line 19: Why are we moving this to the bottom? We are now doing more than needed

ios/WebSocket/WebSocketManager.swift
```diff
@@ -50,33 +45,37 @@ class WebSocketManager: NSObject {
...
+            }
+        }
     }
-    
+
     func getWebSocket(for url:URL) -> WebSocket? {
-        return webSockets[url]
-    }
-    
-    func disconnectAll() -> Void {
```
enahum, line 69: Why this function is being removed?

ios/WebSocket/WebSocketManager.swift
```diff
@@ -50,33 +45,37 @@ class WebSocketManager: NSObject {
...
-        for ws in webSockets {
-            ws.value.disconnect()
-        }
+        return queue.sync { webSockets[url] }
     }
 
     func invalidateClient(for url:URL) -> Void {
-        guard let webSocket = getWebSocket(for: url) else {
+        let webSocket: WebSocket? = queue.sync {
+            return webSockets.removeValue(forKey: url)
```
enahum: The removal should happen at the end when everything succeeds

ios/WebSocket/WebSocketManager.swift
```diff
@@ -88,11 +87,16 @@ class WebSocketManager: NSObject {
...
-        webSockets.removeAll()
+        let snapshot: [WebSocket] = queue.sync {
+            let values = Array(webSockets.values)
+            webSockets.removeAll()
+            return values
+        }
+        for ws in snapshot {
+            ws.disconnect()
+        }
     }
```
enahum, line 108: This is throwing me off, why do you need to make a copy?

enahum (changes requested):

ios/WebSocket/WebSocketManager.swift
Willyfrog: no particular reason other than the call to it was removed and thus it would never be called. 

I guess the code there could have been moved here.

ios/WebSocket/WebSocketManager.swift
Willyfrog: ah, this made sense after a concern from AI review, but not in the general scheme, fixing

ios/WebSocket/WebSocketManager.swift
Willyfrog: when I asked AI about it (I'm basically a swift noob) he replied that it was to avoid the deadlock:

> The copy is needed to avoid a deadlock. ws.disconnect() can trigger Starscream delegate callbacks that call back into the manager (e.g., getWebSocket),
  which would try to acquire the same serial queue via queue.sync. If we iterated inside the lock, that re-entrant queue.sync would deadlock. So we clear the
  dictionary and grab the values inside the lock, then iterate outside it.

Which sounded fine. But if you think this is not needed, I'll be happy to fix it, but I need an expert eye on there

ios/WebSocket/WebSocketManager.swift
```diff
@@ -50,33 +45,37 @@ class WebSocketManager: NSObject {
...
-        webSockets[url] = webSocket
+
+        queue.sync {
+            if let existing = webSockets[url] {
+                existing.delegate = delegate
+                webSocket.delegate = nil
+            } else {
+                webSockets[url] = webSocket
+            }
+        }
```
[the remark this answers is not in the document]
Willyfrog: this is intended

Willyfrog:
@enahum I think I addressed all, can you take a look again?

larkox (approved):

enahum (approved):

PR #164, opened 2026-04-14
Explicitly set MediaType for different body types

ddzobov:
#### Summary
If Mattermost server published behind WAF, then WAF may block requests from Android clients, because by default it not sending any Content-Type headers in POST requests.

okhttp3 not sets Content-Type headers by default, so we need to do this explicitly.

https://github.com/mattermost/mattermost-mobile/issues/9689

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
+++ b/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
@@ -23,6 +23,7 @@ import com.mattermost.networkclient.metrics.RequestMetadata
 import com.mattermost.networkclient.metrics.getNetworkType
 import okhttp3.*
 import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.tls.HandshakeCertificates
 import org.json.JSONArray
 import org.json.JSONObject
@@ -46,6 +47,8 @@ const val CERTIFICATES_PATH = "certs"
 internal class NetworkClient(private val context: Context, private val baseUrl: HttpUrl? = null, options: ReadableMap? = null, cookieJar: CookieJar? = null) {
     private var okHttpClient: OkHttpClient
     private var webSocketUri: URI? = null
+    private val jsonUtf8MediaType = "application/json; charset=utf-8".toMediaType()
+    private val textUtf8MediaType = "text/plain; charset=utf-8".toMediaType()
 
     var clientHeaders: HashMap<String, Any?> = hashMapOf()
     var clientRetryInterceptor: Interceptor? = null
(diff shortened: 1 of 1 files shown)


enahum:
Thanks for the PR @ddzobov as soon as I have a bit of time, I'll check if this in fact resolves the issue with WAF.

But two things:

1. iOS seems to be left out, is it because is currently working or because you only focus on Android? If the latter please do test on iOS and update your PR accordingly
2. Isn't it better to pass the header in the options and include these two types instead of forcing it as this PR does?

ddzobov:
> Thanks for the PR @ddzobov as soon as I have a bit of time, I'll check if this in fact resolves the issue with WAF.
> 
> But two things:
> 
> 1. iOS seems to be left out, is it because is currently working or because you only focus on Android? If the latter please do test on iOS and update your PR accordingly
> 2. Isn't it better to pass the header in the options and include these two types instead of forcing it as this PR does?

1. Yes, iOS code sending application/json now, these users dont have problems at now
2. As i see from code, header passed in options will have bigger priority, so it will work in both cases

enahum:
> > Thanks for the PR @ddzobov as soon as I have a bit of time, I'll check if this in fact resolves the issue with WAF.
> 
> > 
> 
> > But two things:
> 
> > 
> 
> > 1. iOS seems to be left out, is it because is currently working or because you only focus on Android? If the latter please do test on iOS and update your PR accordingly
> 
> > 2. Isn't it better to pass the header in the options and include these two types instead of forcing it as this PR does?
> 
> 
> 
> 1. Yes, iOS code sending application/json now, these users dont have problems at now
> 
> 2. As i see from code, header passed in options will have bigger priority, so it will work in both cases

So if it works in both cases, I rather not do this here and set the default headers on the mobile app instead and not in the library

ddzobov:
> > > Thanks for the PR @ddzobov as soon as I have a bit of time, I'll check if this in fact resolves the issue with WAF.
> > 
> > 
> > > 
> > 
> > 
> > > But two things:
> > 
> > 
> > > 
> > 
> > 
> > > 1. iOS seems to be left out, is it because is currently working or because you only focus on Android? If the latter please do test on iOS and update your PR accordingly
> > 
> > 
> > > 2. Isn't it better to pass the header in the options and include these two types instead of forcing it as this PR does?
> > 
> > 
> > 
> > 1. Yes, iOS code sending application/json now, these users dont have problems at now
> > 2. As i see from code, header passed in options will have bigger priority, so it will work in both cases
> 
> So if it works in both cases, I rather not do this here and set the default headers on the mobile app instead and not in the library

Yes, but current logic in iOS code not sets application/json header explicitly too, currently it determines header implicitly based on serialization type, so with changes in this PR logic will be equal.

If we will set headers on the mobile app, we need to do more complex changes in each request in each class call for request.

ddzobov:
Hello
Any news?

PR #171, opened 2026-07-17
feat: [MM-69203] Implement collection, storage and attachment of session attributes

devinbinnie:
#### Summary
DRAFT: Implement collection, storage and attachment of session attributes

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69203

### DIFF
--- a/android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
+++ b/android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
@@ -14,6 +14,7 @@ import com.facebook.react.bridge.WritableMap
 import com.facebook.react.modules.network.ForwardingCookieHandler
 import com.facebook.react.modules.network.ReactCookieJarContainer
 import com.mattermost.networkclient.helpers.KeyStoreHelper
+import com.mattermost.networkclient.sessionattributes.SessionAttributesEngine
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.runBlocking
 import okhttp3.Call
@@ -358,6 +359,34 @@ class ApiClientModuleImpl(appContext: Context) {
         } catch (error: Exception) {
             return promise.reject(error)
         }
+    }
+
+    fun setSessionAttributesEnabled(serverUrl: String, enabled: Boolean) {
+        SessionAttributesEngine.getInstance(context).setEnabled(serverUrl, enabled)
+    }
+
+    fun removeSessionAttributesServer(serverUrl: String) {
+        SessionAttributesEngine.getInstance(context).removeServer(serverUrl)
+    }
+
+    fun setSessionAttributesManifest(serverUrl: String, manifest: String) {
+        SessionAttributesEngine.getInstance(context).setManifest(serverUrl, manifest)
+    }
+
+    fun upsertSessionAttributesField(serverUrl: String, field: String) {
+        SessionAttributesEngine.getInstance(context).upsertManifestField(serverUrl, field)
+    }
+
+    fun removeSessionAttributesField(serverUrl: String, name: String) {
+        SessionAttributesEngine.getInstance(context).removeManifestField(serverUrl, name)
+    }
+
+    fun setSessionAttributesStableValues(values: String) {
+        SessionAttributesEngine.getInstance(context).setStableValues(values)
+    }
+
+    fun getSessionAttributesHeader(serverUrl: String): String? {
+        return SessionAttributesEngine.getInstance(context).getOutboundHeader(serverUrl)
     }
 
     // Methods to use with native implementations
--- a/android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
+++ b/android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
@@ -133,6 +133,7 @@ class ApiClientModuleImpl(appContext: Context) {
 
     init {
         setCtx(appContext)
+        SessionAttributesEngine.init(appContext)
         migrateSharedPreferences(appContext)
         setCookieJar(appContext)
     }
@@ -362,31 +363,31 @@ class ApiClientModuleImpl(appContext: Context) {
     }
 
     fun setSessionAttributesEnabled(serverUrl: String, enabled: Boolean) {
-        SessionAttributesEngine.getInstance(context).setEnabled(serverUrl, enabled)
+        SessionAttributesEngine.setEnabled(serverUrl, enabled)
     }
 
     fun removeSessionAttributesServer(serverUrl: String) {
-        SessionAttributesEngine.getInstance(context).removeServer(serverUrl)
+        SessionAttributesEngine.removeServer(serverUrl)
     }
 
     fun setSessionAttributesManifest(serverUrl: String, manifest: String) {
-        SessionAttributesEngine.getInstance(context).setManifest(serverUrl, manifest)
+        SessionAttributesEngine.setManifest(serverUrl, manifest)
     }
 
     fun upsertSessionAttributesField(serverUrl: String, field: String) {
-        SessionAttributesEngine.getInstance(context).upsertManifestField(serverUrl, field)
+        SessionAttributesEngine.upsertManifestField(serverUrl, field)
     }
 
     fun removeSessionAttributesField(serverUrl: String, name: String) {
-        SessionAttributesEngine.getInstance(context).removeManifestField(serverUrl, name)
+        SessionAttributesEngine.removeManifestField(serverUrl, name)
     }
 
     fun setSessionAttributesStableValues(values: String) {
-        SessionAttributesEngine.getInstance(context).setStableValues(values)
+        SessionAttributesEngine.setStableValues(values)
     }
 
     fun getSessionAttributesHeader(serverUrl: String): String? {
-        return SessionAttributesEngine.getInstance(context).getOutboundHeader(serverUrl)
+        return SessionAttributesEngine.getOutboundHeader(serverUrl)
     }
 
     // Methods to use with native implementations
--- a/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
+++ b/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
@@ -161,6 +161,11 @@ internal class NetworkClient(private val context: Context, private val baseUrl: 
             builder.addInterceptor(bearerTokenInterceptor)
         }
 
+        // Runs after BearerTokenInterceptor so the Authorization header is present when the
+        // guard is evaluated. Also covers adaptRCTRequest() since RCT requests are executed
+        // through this same client's okHttpClient.
+        builder.addInterceptor(SessionAttributesInterceptor(context, baseUrlString))
+
         applyClientSslConfiguration(options)
         configureSsl()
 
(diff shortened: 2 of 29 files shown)


android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
```diff
@@ -360,6 +361,34 @@ class ApiClientModuleImpl(appContext: Context) {
         }
     }
 
+    fun setSessionAttributesEnabled(serverUrl: String, enabled: Boolean) {
+        SessionAttributesEngine.getInstance(context).setEnabled(serverUrl, enabled)
```
enahum: I'm seeing that here you call getInstance every time for these functions.. why nit use the dame pattern as this very file and have the context set once?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mattermost/react-native-network-client",
-  "version": "1.10.3",
+  "version": "1.11.0",
```
enahum: This is bump automatically when we release it, no need to change in the PR

android/src/main/java/com/mattermost/networkclient/interceptors/SessionAttributesInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import android.content.Context
+import com.mattermost.networkclient.sessionattributes.SessionAttributes
+import com.mattermost.networkclient.sessionattributes.SessionAttributesConstants
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+
+class SessionAttributesInterceptor(
+    private val context: Context,
```
enahum: I'm not sure of the context here, I think if you set correctly your files for SessionAttributes etc.. the Context would be set once in SessionAttribute during init and not having to pass it as an argument every time

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesEngine.kt
```diff
@@ -0,0 +1,111 @@
...
+        store.saveState(serverUrl, state)
+
+        return Base64.encodeToString(payload.toString().toByteArray(), Base64.NO_WRAP)
+    }
+
+    companion object {
+        @Volatile
+        private var instance: SessionAttributesEngine? = null
+
+        fun getInstance(context: Context): SessionAttributesEngine {
```
enahum: I believe if you define the context as part of the Companion and then set it on init, you would no need to pass the context every time you need an instance

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesCollector.kt
```diff
@@ -0,0 +1,140 @@
...
+                while (addresses.hasMoreElements()) {
+                    val address = addresses.nextElement()
+                    if (!address.isLoopbackAddress && address is Inet4Address) {
+                        return address.hostAddress ?: ""
+                    }
+                }
+            }
+            ""
+        } catch (_: Exception) {
+            ""
```
enahum, line 124: I get that our fallback is an empty ip address, but do you want to log it, to see why it went wrong ?

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesCollector.kt
```diff
@@ -0,0 +1,140 @@
...
+            val wifiManager = context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager
+            val info = wifiManager.connectionInfo
+            val ssid = info?.ssid?.replace("\"", "") ?: ""
+            if (ssid == "<unknown ssid>" || ssid == "0x" || ssid == "Wi-Fi" || ssid == "WLAN") {
+                ""
+            } else {
+                ssid
+            }
+        } catch (_: Exception) {
+            ""
```
enahum, line 141: same here

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesEngine.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    fun removeServer(serverUrl: String) {
+        store.removeState(serverUrl)
+    }
+
+    fun setManifest(serverUrl: String, manifestJson: String) {
+        val manifest = try {
+            JSONArray(manifestJson)
+        } catch (_: Exception) {
+            null
```
enahum, line 34: log? or is fine to omit?

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesEngine.kt
```diff
@@ -0,0 +1,111 @@
...
+            return
+        }
+        store.setManifest(serverUrl, fields)
+    }
+
+    fun upsertManifestField(serverUrl: String, fieldJson: String) {
+        val field = try {
+            JSONObject(fieldJson)
+        } catch (_: Exception) {
+            return
```
enahum, line 54: log? or is fine to omit?

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesEngine.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    fun removeManifestField(serverUrl: String, name: String) {
+        store.removeField(serverUrl, name)
+    }
+
+    fun setStableValues(valuesJson: String) {
+        val json = try {
+            JSONObject(valuesJson)
+        } catch (_: Exception) {
+            return
```
enahum, line 68: log? or is fine to omit?

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesStore.kt
enahum, line 1: I'm a bit worried about the performance off all of this, not only this is reading and writing encrypted data but there is also a lock, have you considered caching in memory map the deserialized ServerSessionAtriibuteState per server URL, that is populated on read if there is no value and invalidated/updated on write? Also I would be a bit careful having `runBlocking` inside a `synchronized`, this can cause deadlocks if not careful.

in addition, there are several try/catch blocks, is there any benefit on logging the errors ?

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -161,6 +161,11 @@ internal class NetworkClient(private val context: Context, private val baseUrl:
             builder.addInterceptor(bearerTokenInterceptor)
         }
 
+        // Runs after BearerTokenInterceptor so the Authorization header is present when the
+        // guard is evaluated. Also covers adaptRCTRequest() since RCT requests are executed
+        // through this same client's okHttpClient.
+        builder.addInterceptor(SessionAttributesInterceptor(context, baseUrlString))
```
enahum: I think we want to add this one behind a configuration option, this library is used by other people as well, not only internally at MM, I'm not sure this should be enabled for everyone.

ios/patches/apply_patches.rb
```diff
@@ -68,7 +68,12 @@ def patch_additions_present?(file, repo_root, directory_arg)
 
         def apply_patch(file)
             repo_root = `git rev-parse --show-toplevel`.strip
-            directory_arg = Dir.glob(Pathname(repo_root).join("**/**/Pods")).first.sub("#{repo_root}/", "")
+            pods_dir = Dir.glob(Pathname(repo_root).join("**/**/Pods")).first
+            unless pods_dir
+                Pod::UI.puts "Skipping #{file} (Pods directory not found yet)"
+                return
+            end
+            directory_arg = pods_dir.sub("#{repo_root}/", "")
```
enahum: why this change ?

ios/SessionAttributes/SessionAttributes.swift
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import Foundation
+
+/// React-free entry point for native code outside this library (e.g. Gekidou,
+/// app extensions, standalone URLSession/OkHttp usage) to resolve the outbound
```
enahum:
```suggestion
/// app extensions, standalone URLSession usage) to resolve the outbound
```

ios/NetworkClient.swift
```diff
@@ -294,10 +294,9 @@ extension NetworkClient {
             adapters.append(BearerAuthenticationAdapter())
         }
 
-        if (adapters.isEmpty) {
-            return Interceptor(retriers: retriers)
-        }
-    
+        // Must run after BearerAuthenticationAdapter so the Authorization header is present.
+        adapters.append(SessionAttributesAdapter())
```
enahum: adapter should be optional and it should have a configuration option, same as bearer token. This library is used by others not only by MM, they may not want to configure this.

react-native-network-client.podspec
```diff
@@ -12,31 +12,43 @@ Pod::Spec.new do |s|
...
-  else
-    s.pod_target_xcconfig    = {
-      "DEFINES_MODULE" => "YES",
-      "BUILD_LIBRARY_FOR_DISTRIBUTION" => "YES",
-      "OTHER_SWIFT_FLAGS" => "-no-verify-emitted-module-interface"
-    }
+
+  # SessionAttributes lives in its own React-free subspec so it can be linked by
+  # app extensions / standalone native code without pulling in React.
+  s.subspec 'SessionAttributes' do |sa|
```
enahum, line 19: this seems like a valid approach, let's see if in fact does work in our extensions ;)

enahum (changes requested):

android/src/main/java/com/mattermost/networkclient/ApiClientModuleImpl.kt
devinbinnie: Good call, fixed.

android/src/main/java/com/mattermost/networkclient/interceptors/SessionAttributesInterceptor.kt
devinbinnie: Yeah that's a better approach.

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesCollector.kt
devinbinnie: Yeah I'm not sure why it would fail (or if we'll care) but for now we can log so if we see it fail in some cases we'll know why. I'll add the logging where requested.

ios/patches/apply_patches.rb
devinbinnie: Reverted, I thought it was needed but I was wrong.

android/src/main/java/com/mattermost/networkclient/sessionattributes/SessionAttributesStore.kt
devinbinnie: I've added a cache map to ensure we aren't re-updating the store every single time, and I've added some logging when we fail to fetch.

android/src/main/java/com/mattermost/networkclient/NetworkClient.kt
```diff
@@ -496,6 +504,19 @@ internal class NetworkClient(private val context: Context, private val baseUrl:
         return null
     }
 
+    private fun getSessionAttributesInterceptor(options: ReadableMap?): SessionAttributesInterceptor? {
```
enahum, line 507: nice

enahum (approved):
yay!!

larkox (approved):

devinbinnie:
@coderabbitai resolve all and approve. this has been reviewed already.

